PR #8, opened 2020-05-03
Fixing ordering issues by looking up previous node

sarveshnagpal:
- Fixing ordering issues by looking up previous node
- Fit & finish around Clarity Developer Tools


packages/clarity-decode/types/layout.d.ts
```diff
@@ -17,9 +17,10 @@ export import Constant = Layout.Constant;
 export interface DomData {
     id: number;
     parent: number;
-    next: number;
+    previous: number;
     tag: string;
     position: number;
     attributes?: Layout.Attributes;
     value?: string;
+    next?: number; /* deprecated since v0.4.5 */
```
toby-walker, line 25: >next [](start = 4, length = 4)

This isn't backwards compatible, is it? Should this be 0.5 instead?

packages/clarity-devtools/static/manifest.json
```diff
@@ -1,16 +1,23 @@
...
+  "version": "0.4.5",
+  "version_name": "0.4.5",  
+  "minimum_chrome_version": "50",
   "devtools_page": "devtools.html",
+  "icons": {
+    "16": "icon-16.png",
+    "32": "icon-32.png",
+    "48": "icon-48.png",
+    "64": "icon-64.png",
+    "128": "icon-128.png"
```
toby-walker, line 14: Is there a source for these like an SVG or other file? Is it already checked in?

packages/clarity-js/src/layout/dom.ts
```diff
@@ -117,31 +117,28 @@ export function add(node: Node, parent: Node, data: NodeInfo, source: Source): v
...
+            value.parent = parentId;
             // Move this node to the right location under new parent
             if (parentId !== null && parentId >= 0) {
-                if (nextId !== null && nextId >= 0) {
-                    values[parentId].children.splice(nextId + 1, 0 , id);
-                } else {
-                    values[parentId].children.push(id);
-                }
+                let childIndex = previousId !== null ? values[parentId].children.indexOf(previousId) + 1 : values[parentId].children.length;
+                values[parentId].children.splice(childIndex, 0 , id);
```
toby-walker, line 141: >  [](start = 62, length = 1)

Minor nit: Extra space before ,

packages/clarity-js/src/layout/dom.ts
```diff
@@ -117,31 +117,28 @@ export function add(node: Node, parent: Node, data: NodeInfo, source: Source): v
...
+            let oldParentId = value.parent;
+            value.parent = parentId;
             // Move this node to the right location under new parent
             if (parentId !== null && parentId >= 0) {
-                if (nextId !== null && nextId >= 0) {
-                    values[parentId].children.splice(nextId + 1, 0 , id);
-                } else {
-                    values[parentId].children.push(id);
-                }
+                let childIndex = previousId !== null ? values[parentId].children.indexOf(previousId) + 1 : values[parentId].children.length;
```
toby-walker, line 140: >!== [](start = 44, length = 3)

Style question: Do you prefer to write these as above or as: 

previousId === null ? values[values[parentId].children.length] : values[parentId].children.indexOf(previousId) + 1 

So that the condition is not negated? To me the above is my preference as I find it easier to understand but tastes on this point vary.

packages/clarity-js/src/layout/dom.ts
```diff
@@ -117,31 +117,28 @@ export function add(node: Node, parent: Node, data: NodeInfo, source: Source): v
...
         // Handle case where parent might have been updated
-        if (value["parent"] !== parentId) {
+        if (value.parent !== parentId) {
             changed = true;
-            let oldParentId = value["parent"];
-            value["parent"] = parentId;
+            let oldParentId = value.parent;
+            value.parent = parentId;
             // Move this node to the right location under new parent
             if (parentId !== null && parentId >= 0) {
```
toby-walker, line 139: What if the next sibling also changes? What should happen here? Should you not check both?

packages/clarity-js/src/layout/dom.ts
```diff
@@ -367,11 +364,11 @@ function getFullUrl(relative: string): string {
     return a.href;
 }
 
-function getNextId(node: Node): number {
+function getPreviousId(node: Node): number {
     let id = null;
-    while (id === null && node.nextSibling) {
-        id = getId(node.nextSibling);
-        node = node.nextSibling;
+    while (id === null && node.previousSibling) {
```
toby-walker, line 369: What happens when there are nodes witthout ids but are siblings? This skips them. Can you add a comment explaining why that is OK?

packages/clarity-js/types/layout.d.ts
```diff
@@ -58,7 +58,7 @@ export interface NodeInfo {
 export interface NodeValue {
     id: number;
     parent: number;
-    next: number;
+    previous: number;
```
toby-walker, line 61: Definitely not backwards compatible. Needs a major version bump (0.5.0)

packages/clarity-visualize/src/interaction.ts
```diff
@@ -17,32 +17,29 @@ export function scroll(event: Interaction.ScrollEvent, iframe: HTMLIFrameElement
...
+            iframe.removeAttribute("style");
+            iframe.style.position = "relative";
+            iframe.style.width = width + px;
+            iframe.style.height = height + px;
+            iframe.style.transformOrigin = "0 0 0";
+            iframe.style.transform = "scale(" + scale + ")";
+            iframe.style.border = "1px solid #cccccc";
+            iframe.style.overflow = "hidden";
+            iframe.style.left = ((container.clientWidth - (width * scale)) / 2) + px;
+        }  catch { /* do nothing */ }
```
toby-walker, line 42: When can you get an exception? Add comment.

packages/clarity-visualize/src/layout.ts
```diff
@@ -41,7 +41,18 @@ export function markup(event: Layout.DomEvent, iframe: HTMLIFrameElement): void
     let doc = iframe.contentDocument;
     for (let node of data) {
         let parent = element(node.parent);
-        let next = element(node.next);
+        let pivot = element(node.previous);
+        let insert = insertAfter;
+
+        // For backward compatibility. Until v0.4.4 we used the next element to determine right ordering within DOM
+        // Starting with v0.4.5, we moved away from the next element and instead starting sending previous element
```
toby-walker, line 48: Why does this change (using previous instead of next) help? Add that to comment here.

packages/clarity-visualize/src/layout.ts
```diff
@@ -41,7 +41,18 @@ export function markup(event: Layout.DomEvent, iframe: HTMLIFrameElement): void
...
+        // This change fixes several edge cases where positioning of DOM could get inconsistent.
+        // In future, we can get rid of following code, but keeping it for now to ensure backward compatibility.
+        if ("next" in node) {
```
toby-walker, line 51: Question: Would it be better to explicitly check the version of the data you are receiving rather than checking for a field? This is a judgment call but I have a slight preference for explicitly including the version in the logic here so that there is no ambiguity and no potential future problem.

packages/clarity-decode/types/layout.d.ts
sarveshnagpal: It is backward compatible.

packages/clarity-js/src/layout/dom.ts
sarveshnagpal: Thanks, I hate those too. Fixed :)

packages/clarity-js/src/layout/dom.ts
sarveshnagpal: Your recommendation looks good, made the change.

packages/clarity-js/src/layout/dom.ts
sarveshnagpal: If we are pivoting on previous element - we don't need to worry about next sibling since it will have no impact on the current node. Any changes to next sibling will be handled as part of a separate call to this update function.

packages/clarity-js/src/layout/dom.ts
sarveshnagpal: Yes adding a comment. But these nodes are the ones that we skip while walking the DOM like comment tags and script tags.

packages/clarity-js/types/layout.d.ts
sarveshnagpal: It is backward compatible. This is an internal detail to decode & js that by design are supposed to be tightly coupled. No impact to any other layer consuming this data. Happy to discuss offline if we should make a major version bump or not.

packages/clarity-visualize/src/interaction.ts
sarveshnagpal: Sure - iframe may or may not have access to the parent page. I will add a comment.

packages/clarity-visualize/src/layout.ts
sarveshnagpal: I have a slight preference for feature detection over version checking but in this case even if we wanted to enforce version check, we will need to change the interface for visualization and that will turn into a bigger change. In future I do want to update the interface such that we can't call replay without passing in the version. Today it's possible to send in events without ever sending the envelope. I think it's not great but want to address that shortcoming as part of a different change.

packages/clarity-visualize/src/layout.ts
sarveshnagpal: Sure, thanks.

PR #14, opened 2020-05-15
Adding the ability to filter out pages with missing layout

sarveshnagpal:


packages/clarity-js/src/data/upload.ts
```diff
@@ -77,8 +77,9 @@ export function queue(data: Token[]): void {
                 } else { metric.count(Metric.LayoutBytes, event.length); }
                 break;
             case Event.BoxModel:
+                metric.count(Metric.RegionBytes, event.length);
+                break;
             case Event.Document:
-                metric.count(Metric.LayoutBytes, event.length);
```
toby-walker, line 81: Please add a comment why this case doesn't count for anything. It's less obvious than some of the others.

sarveshnagpal: Done

PR #18, opened 2020-06-05
Refactor code to make things more efficient

sarveshnagpal:
- Using shorter ids for userId / sessionId
- Changing pageId to be a sequence number
- Drop Event.Target & Event.Hash
- Drop some internal metrics
- Simplify performance module
- Add support for Event.Baseline to keep state across payloads
- Include regionId with every interaction
- Extend click event to include relative (x,y) and element hash
- Drop Event.Page and introduce "Feature" for string values.
- Renamed occurrences of "boxmodel" => region to stay consistent
- Fix race condition with click events


packages/clarity-js/src/core/measure.ts
```diff
@@ -8,11 +8,11 @@ export default function(method: Function): Function {
         let start = performance.now();
         method.apply(this, arguments);
         let duration = performance.now() - start;
-        metric.accumulate(Metric.TotalDuration, duration);
+        metric.sum(Metric.TotalCost, duration);
         metric.count(Metric.InvokeCount);
```
swaathee, line 12: what are we trying to measure from invoke count? - if it is the number of times the clarity yielded, this shouldnt be counted here since measure is called for multiple function calls

swaathee:
>         restart(method);

this gets reset in the next line 

---
Refers to: packages/clarity-js/src/core/task.ts:110 in d36860e. [](commit_id = d36860e58a9151ec33d4a1083bef1f72097307f9, deletion_comment = False)

packages/clarity-js/src/data/index.ts
```diff
@@ -12,9 +13,10 @@ export { upgrade } from "@src/data/upgrade";
 
 export function start(): void {
     metric.start();
+    dimension.start();
     measure(upload.start)();
```
swaathee, line 17: why is this not measured

packages/clarity-js/src/data/metric.ts
```diff
@@ -2,44 +2,45 @@ import { Event, Metric, MetricData } from "@clarity-types/data";
...
 }
 
 export function max(metric: Metric, value: number): void {
     if (!(metric in data)) { data[metric] = 0; }
-    data[metric] = Math.max(value, data[metric]);
-    track(metric);
+    if (value > data[metric] || data[metric] === 0) {
+        if (!(metric in updates)) { updates[metric] = 0; }
+        updates[metric] += value - data[metric];
+        data[metric] = value;
```
swaathee, line 36: why? we can do max of all the max values that we are sending - sending incremental value for max makes it unclear

packages/clarity-js/src/data/dimension.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export function end(): void {
+    data = {};
+    updates = {};
+}
+
+export function log(dimension: Dimension, value: string): void {
+    if (!(dimension in data)) { data[dimension] = []; }
+    if (data[dimension].indexOf(value) < 0) {
+        data[dimension].push(value);
```
swaathee, line 20: what if  a dimension gets updated?

swaathee:
>             tokens.push(script.data.message);

if we are only going to track script error count - should we just send details for clarity script errors 

---
Refers to: packages/clarity-js/src/diagnostic/encode.ts:14 in e447c85. [](commit_id = e447c8511edff8bd3ca47faa2537138c5d9ba941, deletion_comment = False)

packages/clarity-js/src/layout/dom.ts
```diff
@@ -50,15 +49,10 @@ export function extractRegions(root: ParentNode): void {
     for (let key in config.regions) {
         // We check for regions in the beginning (document) and later whenever there are new additions or modifications to DOM (mutations)
         // Since mutations may happen on leaf nodes too, e.g. textnodes, which may not support all selector APIs.
```
swaathee, line 51: this gets called before dom discovery - is it possible that between this and when mutation observer is called (there could be other elements that are added and not tracked as a part of region?)

packages/clarity-js/src/layout/dom.ts
```diff
@@ -80,12 +74,12 @@ export function add(node: Node, parent: Node, data: NodeInfo, source: Source): v
     let previousId = getPreviousId(node);
     let masked = true;
     let parentValue = null;
-    let region = regionMap.has(node) ? regionMap.get(node) : null;
+    let regionId = regionMap.has(node) ? getId(node) : null;
```
swaathee, line 77: this seems wrong - how is this different from node id?

packages/clarity-js/src/layout/dom.ts
```diff
@@ -140,7 +134,7 @@ export function update(node: Node, parent: Node, data: NodeInfo, source: Source)
                 let childIndex = previousId === null ? values[parentId].children.length : values[parentId].children.indexOf(previousId) + 1;
                 values[parentId].children.splice(childIndex, 0, id);
```
swaathee, line 135: will this work if an element is inserted as a first child? - it seems like we will consider it inserted as last child?

packages/clarity-js/src/interaction/baseline.ts
```diff
@@ -0,0 +1,49 @@
...
+    update = false;
+}
+
+export function visibility(visible: string): void {
+    buffer.visible = visible === "visible" ? BooleanFlag.True : BooleanFlag.False;
+    update = true;
+}
+
+export function track(event: Event, x: number, y: number): void {
+    switch (event) {
```
swaathee, line 26: can you explain how this helps in  sending previous state?
i.e - from my understanding - we are tracking changes to pointer and sending the last state as a part of the current upload, i don't see how we are resending it in the next payload?

packages/clarity-js/src/performance/observer.ts
```diff
@@ -48,28 +45,34 @@ function handle(entries: PerformanceObserverEntryList): void {
...
-                case "paint":
-                    paint.compute(entry);
+                    dimension.log(Dimension.NetworkHosts, host(entry.name));
                     break;
                 case "longtask":
-                    longtask.compute(entry as LongTaskEntry);
+                    metric.count(Metric.LongTaskCount);
+                    break;
+                case "first-input":
+                    if (visible) { metric.count(Metric.FirstInputDelay, entry["processingStart"] - entry.startTime); }
```
swaathee, line 62: >count [](start = 42, length = 5)

this feel weird - should this be max or sum? (i know it doesn't make a difference since we would only get this once per page but count seems like it is a incremental thing)

packages/clarity-js/src/interaction/encode.ts
```diff
@@ -28,26 +26,37 @@ export default async function(type: Event): Promise<void> {
...
+                let pTarget = metadata(entry.data.target as Node);
                 tokens = [entry.time, entry.event];
-                tokens.push(observe(entry.data.target as TargetInfo));
+                tokens.push(pTarget.id);
                 tokens.push(entry.data.x);
                 tokens.push(entry.data.y);
+                if (pTarget.region) { tokens.push(pTarget.region); }
                 queue(tokens);
+                baseline.track(entry.event, entry.data.x, entry.data.y); // Track changes to baseline
             }
```
swaathee, line 38: since region is a id - how do you map to region name for clicks while being stateless?

packages/clarity-devtools/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clarity-devtools",
-  "version": "0.5.9",
+  "version": "0.6.0-beta.1",
   "private": true,
```
swaathee, line 4: what is the difference between this and the one in clarity analytics?

packages/clarity-js/src/interaction/click.ts
```diff
@@ -34,22 +36,30 @@ function handler(event: Event, root: Node, evt: MouseEvent): void {
     // If present, we use the returned link element to populate text and link properties below
     let a = link(t);
 
+    // Get layout rectangle for the target element
+    let l = layout(t as Element);
+    let eX = l ? x - l.x : 0;
+    let eY = l ? y - l.y : 0;
```
swaathee, line 42: divide by total width?

packages/clarity-js/src/data/metadata.ts
```diff
@@ -1,57 +1,45 @@
...
-    let sessionId = session(ts);
-    let pageId = guid();
-    let ua = navigator && "userAgent" in navigator ? navigator.userAgent : "";
-    let upload = Upload.Async;
-    let lean = config.lean ? BooleanFlag.True : BooleanFlag.False;
-    let title = document && document.title ? document.title : null;
-    let e: Envelope = { sequence: 0, version, pageId, userId, sessionId, projectId, upload, end: BooleanFlag.False };
-    let p: PageData = { timestamp: ts, ua, url: location.href, referrer: document.referrer, lean, title };
+    const ts = Math.round(Date.now()); // ensuring that the output of Date.now() is an integer
+    const ua = navigator && "userAgent" in navigator ? navigator.userAgent : ""
```
swaathee, line 13: >"" [](start = 77, length = 2)

nit: keep use of ; consistent

swaathee:
>                 }

how does this update the elements which had current element as a previousSibling?
 Does mutation observer trigger mutations on elements that have sibilings changed (documentation only specified childlist so i am not sure about this)

---
Refers to: packages/clarity-js/src/layout/dom.ts:148 in e447c85. [](commit_id = e447c8511edff8bd3ca47faa2537138c5d9ba941, deletion_comment = False)

packages/clarity-js/src/layout/mutation.ts
```diff
@@ -29,14 +36,19 @@ export function start(): void {
...
-      let value = insertRule.call(this, rule, index);
-      generate(this.ownerNode, Constant.CHARACTER_DATA);
-      return value;
+      try {
+        let value = insertRule.call(this, rule, index);
+        schedule(this.ownerNode);
+        return value;
+      } catch (error) {
+        log.log(Code.CssRules, error, Severity.Info);
+        return null;
```
dan-kez, line 45: Swallowing this error is causing issues per the details documented in this PR https://github.com/microsoft/clarity/pull/62 

Updating your beta branch that is shipped to customers to not swallow this error would be highly appreciated.

PR #19, opened 2020-06-09
Track Region for clicks and other bug fixes 

swaathee:
Changes include
1. Bug fix for invoke count, task yield time,  relative position for clicks 
2. Track region info for clicks and Remove region for pointer and scroll events

### DIFF
--- a/packages/clarity-js/src/interaction/click.ts
+++ b/packages/clarity-js/src/interaction/click.ts
@@ -10,6 +10,7 @@ import { link, target } from "@src/layout/target";
 import encode from "./encode";
 
 export let state: ClickState[] = [];
+let clickPrecision = Math.pow(2, 16);
 
 export function start(): void {
     reset();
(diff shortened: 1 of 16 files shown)


packages/clarity-js/src/layout/region.ts
```diff
@@ -1,40 +1,27 @@
...
-    task.start(timer);
     let values = dom.regions();
     let viewport = getViewport();
 
     for (let value of values) {
-        if (task.shouldYield(timer)) {
-            await task.suspend(timer);
-            viewport = getViewport();
-        }
+        viewport = getViewport();
```
sarveshnagpal, line 20: This is not needed if it's sync. We can use the value above. It was only needed it if was async.

packages/clarity-js/src/interaction/click.ts
```diff
@@ -10,6 +10,7 @@ import { link, target } from "@src/layout/target";
 import encode from "./encode";
 
 export let state: ClickState[] = [];
+let clickPrecision = Math.pow(2, 16);
```
sarveshnagpal, line 13: nitpick: Make it const

PR #20, opened 2020-06-11
Adding support for Timeline event and refactoring Baseline event

sarveshnagpal:
- Adding a timeline event that maintains state for configured timeout
- Refactoring baseline event to include viewport dimensions


packages/clarity-js/src/data/baseline.ts
```diff
@@ -14,33 +14,29 @@ export function start(): void {
...
+        case Event.Resize:
+            buffer.screenWidth = width;
+            buffer.screenHeight = height;
             break;
     }
     update = true;
 }
 
+export function visibility(visible: string): void {
+    buffer.visible = visible === "visible" ? BooleanFlag.True : BooleanFlag.False;
```
swaathee, line 36: we would need time for last visibility state - otherwise we can't compute the visbile duration

packages/clarity-js/src/interaction/timeline.ts
```diff
@@ -0,0 +1,44 @@
...
+        event: Event.Timeline,
+        data: { type: event, target, x, y }
+    });
+}
+
+export function compute(): void {
+    updates = [];
+    let max = envelope.data.start + envelope.data.duration;
+    let min = Math.max(max - config.timeline, 0);
+
```
swaathee, line 31: should this be Math.max(envelope.data.start - config.timeline, 0) ?

packages/clarity-js/src/interaction/timeline.ts
```diff
@@ -0,0 +1,44 @@
...
+        time,
+        event: Event.Timeline,
+        data: { type: event, target, x, y }
+    });
+}
+
+export function compute(): void {
+    updates = [];
+    let max = envelope.data.start + envelope.data.duration;
+    let min = Math.max(max - config.timeline, 0);
```
swaathee, line 30: should max be start of envelope - and min be start of envelope - timeline?

packages/clarity-js/src/interaction/timeline.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    for (let s of state) { 
+        if (s.time >= min && s.time <= max) {
+            updates.push(s);
```
swaathee, line 34: should we think about dropping events that are less than min - if there are too many pointer events, we shouldn't be scanning  a big array everytime

packages/clarity-js/src/interaction/encode.ts
```diff
@@ -53,14 +56,15 @@ export default async function (type: Event): Promise<void> {
                 tokens.push(cTarget.hash);
                 if (cTarget.region) { tokens.push(cTarget.region); }
                 queue(tokens);
-                baseline.track(entry.event, entry.data.x, entry.data.y); // Track changes to baseline
+                if (last) { timeline.track(entry.time, entry.event, cTarget.id, entry.data.x, entry.data.y); }
             }
```
swaathee, line 60: if we are only tracking the last event - how would we compute rage clicks?

packages/clarity-js/src/interaction/timeline.ts
swaathee, line 30: this seems like we are sending data for current payload and not previous payload

packages/clarity-decode/src/interaction.ts
```diff
@@ -72,6 +61,14 @@ export function decode(tokens: Data.Token[]): InteractionEvent {
                 y: tokens[4] as number,
             };
             return { time, event, data: scrollData };
+        case Data.Event.Timeline:
+            let timelineData: Interaction.TimelineData = {
+                type: tokens[2] as number,
+                target: tokens[3] as number,
```
toby-walker, line 67: How would we represent things like click text, clicked link in this model?

packages/clarity-js/src/interaction/timeline.ts
sarveshnagpal: Good point. Will implement.

packages/clarity-js/src/interaction/encode.ts
swaathee, line 60: track for all clicks, not the last only

packages/clarity-js/src/interaction/timeline.ts
sarveshnagpal: Clarified offline.

packages/clarity-js/src/data/baseline.ts
sarveshnagpal: Clarified offline.

sarveshnagpal:
> What events will go into the the timeline vs. other events?

Starting small with only Click, Scroll & Pointer. We will expand as needed.

packages/clarity-js/src/interaction/encode.ts
sarveshnagpal: Fixed, thanks.

packages/clarity-decode/src/interaction.ts
sarveshnagpal: It will still be part of the normal event. This only captures the summary.

PR #22, opened 2020-06-13
add timeline for mutations

swaathee:


packages/clarity-js/src/layout/mutation.ts
```diff
@@ -81,54 +81,55 @@ export function end(): void {
 }
 
 function handle(m: MutationRecord[]): void {
+  console.log(time());
   // Queue up mutation records for asynchronous processing
```
swaathee, line 85: will remove this

packages/clarity-js/src/layout/encode.ts
```diff
@@ -69,7 +71,7 @@ export default async function (type: Event, ts: number = null): Promise<void> {
                 }
                 tokens = tokenize(tokens, metadata);
             }
-
+            if (type == Event.Mutation) { timeline.track(eventTime, type); }
```
sarveshnagpal, line 74: We likely don't need to track this as an entry in Timeline. Can we instead track it inside baseline event? We can discuss this more offline - easier than on PR feedback.

sarveshnagpal: And if we don't log it in timeline, timeline can probably stay where it is within Interactions with non-null x & y. It will be specific to interactions.

packages/clarity-js/src/layout/mutation.ts
```diff
@@ -81,54 +81,55 @@ export function end(): void {
...
-  mutations.push({ time: time(), mutations: m});
+  mutations.push({ time: time(), mutations: m });
   task.schedule(process, Priority.High).then((): void => {
-      measure(doc.compute)();
-      measure(region.compute)();
+    measure(doc.compute)();
```
sarveshnagpal, line 88: Nitpick: We should get our IDE in sync so we don't have to look through these weird space differences.

packages/clarity-js/types/data.d.ts
```diff
@@ -146,6 +146,20 @@ export interface BaselineData {
     screenHeight: number;
 }
 
+export interface TimelineState {
```
sarveshnagpal, line 149: These changes may not be necessary if we don't need to include Mutation as part of the timeline.

packages/clarity-decode/src/interaction.ts
```diff
@@ -67,6 +67,8 @@ export function decode(tokens: Data.Token[]): InteractionEvent {
                 target: tokens[3] as number,
                 x: tokens[4] as number,
                 y: tokens[5] as number,
+                hash: tokens.length > 6 ? tokens[6] as number : null,
```
sarveshnagpal, line 70: Let's drop it until we need it for some case later.

packages/clarity-js/src/data/baseline.ts
```diff
@@ -28,12 +28,16 @@ export function track(event: Event, width: number, height: number): void {
             buffer.screenWidth = width;
             buffer.screenHeight = height;
             break;
+        case Event.Mutation:
+            buffer.activityTime = time;
```
sarveshnagpal, line 32: Let's do a separate function for activity to not mix up the function signature here.

packages/clarity-js/src/data/baseline.ts
```diff
@@ -8,14 +8,14 @@ let buffer: BaselineData = null;
...
 export function start(): void {
+    update = false;
     reset();
 }
 
 export function reset(): void {
     // Baseline state holds the previous values - if it is updated in the current payload, reset the state to current value after sending the previous state
     state = update ? { time: time(), event: Event.Baseline, data: buffer } : state;
-    buffer = buffer ? buffer : { visible: BooleanFlag.True, docWidth: 0, docHeight: 0, screenWidth: 0, screenHeight: 0 };
-    update = false;
```
sarveshnagpal, line 18: Removing update here will mean that we are sending baseline every time. Is that a desired thing? Earlier we were only sending when there was an update.

packages/clarity-js/src/interaction/timeline.ts
```diff
@@ -16,11 +16,17 @@ export function reset(): void {
     updates = [];
 }
 
-export function track(time: number, event: Event, target:number, x: number, y: number): void {
+export function track(time: number, event: Event, target: number, x: number, y: number, region?: number): void {
```
sarveshnagpal: Should region be a number? We won't have enough information to decode number to string in this case. Something to think about here.

PR #23, opened 2020-06-15
Add the ability to react to actions from upload endpoint

sarveshnagpal:
Current supported actions: end & upgrade


packages/clarity-js/src/clarity.ts
```diff
@@ -105,7 +104,7 @@ export function tag(key: string, value: string): void {
 
 export function upgrade(key: string): void {
   // Do not process upgrade call if Clarity is not already activated and in lean mode
```
swaathee, line 106: >and in lean mode [](start = 69, length = 16)

nit : remove comment

sarveshnagpal: Thanks will do.

PR #24, opened 2020-06-18
Rethinking tags and adding support to set custom variables

sarveshnagpal:


packages/clarity-js/src/data/variable.ts
```diff
@@ -0,0 +1,35 @@
+import { Constant, Event, VariableData } from "@clarity-types/data";
+import encode from "./encode";
+
+export let data: VariableData = null;
+
+export function set(variable: string, value: string): void {
+    log(variable, value);
+    encode(Event.Variable);
+}
```
swaathee, line 9: do not encode individual calls - and instead use compute below for sending a single event per payload before upload. same for tags

packages/clarity-decode/src/data.ts
```diff
@@ -45,6 +45,13 @@ export function decode(tokens: Data.Token[]): DataEvent {
                 activityTime: tokens[11] as number
             }
             return { time, event, data: baselineData };
+        case Data.Event.Variable:
+            let v = 2; // Start from 3rd index since first two are used for time & event
+            let variables: Data.VariableData = {};
+            while (v < tokens.length) {
```
swaathee, line 51: >tokens.length [](start = 23, length = 13)

should this be tokens.length + 1

packages/clarity-decode/src/data.ts
```diff
@@ -9,7 +9,7 @@ export function decode(tokens: Data.Token[]): DataEvent {
             let ping: Data.PingData = { gap: tokens[2] as number };
             return { time, event, data: ping };
         case Data.Event.Tag:
-            let tag: Data.TagData = { key: tokens[2] as string, value: tokens[3] as string[] };
+            let tag: Data.TagData = { tag: tokens[2] as string };
```
toby-walker, line 12: Why no value any more?

packages/clarity-js/src/clarity.ts
```diff
@@ -95,10 +95,24 @@ export function end(): void {
   }
 }
 
-export function tag(key: string, value: string): void {
+export function tag(name: string): void {
```
toby-walker, line 98: So tags are now more like, well, normal tags that accumulate for a page? What if the same tag is used multiple times?

toby-walker:
How do the new tag and variable show up in the analytics data?

packages/clarity-js/src/data/variable.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export function identify(userId: string, sessionId: string = null, pageId: string = null): void {
+    log(Constant.RESERVED_USER_ID_VARIABLE, userId);
```
toby-walker, line 12: Do we care if the user sets these directly instead of calling this?

packages/clarity-js/src/data/variable.ts
```diff
@@ -0,0 +1,35 @@
+import { Constant, Event, VariableData } from "@clarity-types/data";
+import encode from "./encode";
+
+export let data: VariableData = null;
+
+export function set(variable: string, value: string): void {
```
toby-walker, line 6: What happens if the same variable is set multiple times? What does that mean if it happens?

toby-walker:
Are you planning toadd custom events as part of this PR?

packages/clarity-js/src/data/variable.ts
sarveshnagpal: For tags we do want to encode individual calls so we send one tag per event. That's because time of the tag matters. In case of variables you are right and I've made the change.

packages/clarity-decode/src/data.ts
sarveshnagpal: May be I am missing something here... but what we have looks right. If I add +1 then we will attempt to read tokens[tokens.length] which will be undefined.

packages/clarity-decode/src/data.ts
sarveshnagpal: Like you said tags are not like normal tags :)

packages/clarity-js/src/clarity.ts
sarveshnagpal: We capture them in the events and we don't de-dupe at collection time. De-duping may happen on server but that's outside the scope of client instrumentation.

packages/clarity-js/src/data/variable.ts
sarveshnagpal: No. Not worth adding extra code for that in my opinion. Whenever we work on the documentation, this can be included there.

packages/clarity-js/src/data/variable.ts
sarveshnagpal: Since these are variables, the assumption is that the value will be overridden and the last value will prevail.

packages/clarity-decode/src/data.ts
toby-walker: to me too.

---
In reply to: [443007284](https://github.com/microsoft/clarity/pull/24#discussion_r443007284) [](ancestors = 443007284)

packages/clarity-js/src/clarity.ts
```diff
@@ -79,7 +79,7 @@ export function pause(): void {
 export function resume(): void {
   if (active) {
     task.resume();
-    tag(Constant.CLARITY, "resume");
+    event(Constant.CLARITY, Constant.RESUME);
```
toby-walker, line 82: >Constant [](start = 10, length = 8)

This makes a lot more sense now.

packages/clarity-js/types/data.d.ts
```diff
@@ -62,7 +66,10 @@ export const enum Dimension {
     Referrer = 2,
     PageTitle = 3,
     NetworkHosts = 4,
-    Tags = 5
+    SchemaType = 5,
+    ProductBrand = 6,
```
toby-walker, line 70: Where do these come from in the code?

packages/clarity-decode/src/data.ts
swaathee: sorry i meant -1 since we are incrementing twice inside the loop 

---
In reply to: [444364988](https://github.com/microsoft/clarity/pull/24#discussion_r444364988) [](ancestors = 444364988,443007284)

packages/clarity-js/types/data.d.ts
sarveshnagpal: layout/node.ts

PR #25, opened 2020-06-29
Tslint microsoft contrib

swaathee:


lerna.json
```diff
@@ -2,7 +2,7 @@
   "packages": [
     "packages/*"
   ],
-  "version": "0.5.9",
+  "version": "0.6.0",
```
sarveshnagpal, line 5: 0.5.10 please - not a minor version update.

packages/clarity-decode/src/global.ts
```diff
@@ -1,6 +1,7 @@
 import * as decode from "@src/clarity";
 
 // Expose clarity variable globally to allow access to public interface in a browser
+// tslint:disable-next-line:no-typeof-undefined
 if (typeof window !== "undefined") {
```
sarveshnagpal, line 5: if (window) is probably ok here.

PR #31, opened 2020-08-07
Separating events for playback & analysis

sarveshnagpal:
Also includes:
- Configuration to read custom cookies
- Upgrade the whole session instead of a single page
- Bumping up the package version

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -2,7 +2,7 @@
   "packages": [
     "packages/*"
   ],
-  "version": "0.6.0-beta.4",
+  "version": "0.6.0-beta.5",
   "npmClient": "yarn",
   "useWorkspaces": true
 }
--- a/packages/clarity-decode/src/clarity.ts
+++ b/packages/clarity-decode/src/clarity.ts
@@ -18,8 +18,10 @@ export function decode(input: string): DecodedPayload {
     let envelope = data.envelope(json.e);
     let timestamp = Date.now();
     let payload: DecodedPayload = { timestamp, envelope };
+
     // Sort encoded events by time to simplify summary computation
-    let encoded: Data.Token[][] = json.d.sort((a: Data.Token[], b: Data.Token[]): number => (a[0] as number) - (b[0] as number));
+    let encoded: Data.Token[][] = json.p ? json.a.concat(json.p) : json.a;
+    encoded = encoded.sort((a: Data.Token[], b: Data.Token[]): number => (a[0] as number) - (b[0] as number));
 
     // Check if the incoming version is compatible with the current running code
     // We do an exact match for the major version and minor version.
--- a/packages/clarity-decode/src/clarity.ts
+++ b/packages/clarity-decode/src/clarity.ts
@@ -20,6 +20,9 @@ export function decode(input: string): DecodedPayload {
     let payload: DecodedPayload = { timestamp, envelope };
 
     // Sort encoded events by time to simplify summary computation
+    // It's possible for individual events to be out of order, dependent on how they are buffered on the client
+    // E.g. scroll events are queued internally before they are sent over the wire.
+    // By comparison, events like resize & click are sent out immediately.
     let encoded: Data.Token[][] = json.p ? json.a.concat(json.p) : json.a;
     encoded = encoded.sort((a: Data.Token[], b: Data.Token[]): number => (a[0] as number) - (b[0] as number));
 
--- a/packages/clarity-js/src/data/metadata.ts
+++ b/packages/clarity-js/src/data/metadata.ts
@@ -3,6 +3,7 @@ import config from "@src/core/config";
 import * as dimension from "@src/data/dimension";
 import * as metric from "@src/data/metric";
 import hash from "@src/data/hash";
+import { set } from "@src/data/variable";
 
 export let data: Metadata = null;
 
@@ -27,6 +28,12 @@ export function start(): void {
   // Metrics
   metric.max(Metric.ClientTimestamp, ts);
   metric.max(Metric.Playback, config.lean ? BooleanFlag.False : BooleanFlag.True);
+
+  // Read cookies specified in configuration
+  for (let key of config.cookies) {
+    let value = cookie(key);
+    if (value) { set(key, value); }
+  }
 
   // Track ids using a cookie if configuration allows it
   track();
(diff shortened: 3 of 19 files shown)


lerna.json
```diff
@@ -2,7 +2,7 @@
   "packages": [
     "packages/*"
   ],
-  "version": "0.6.0-beta.4",
+  "version": "0.6.0-beta.5",
```
toby-walker, line 5: Very minor but seems like a big enough update to make this 0..6.1

packages/clarity-decode/src/clarity.ts
```diff
@@ -18,8 +18,10 @@ export function decode(input: string): DecodedPayload {
     let envelope = data.envelope(json.e);
     let timestamp = Date.now();
     let payload: DecodedPayload = { timestamp, envelope };
+
     // Sort encoded events by time to simplify summary computation
-    let encoded: Data.Token[][] = json.d.sort((a: Data.Token[], b: Data.Token[]): number => (a[0] as number) - (b[0] as number));
+    let encoded: Data.Token[][] = json.p ? json.a.concat(json.p) : json.a;
+    encoded = encoded.sort((a: Data.Token[], b: Data.Token[]): number => (a[0] as number) - (b[0] as number));
```
toby-walker, line 24: Minor: Why do you need to sort them? Maybe add comment.

packages/clarity-js/src/data/metadata.ts
```diff
@@ -28,6 +32,12 @@ export function start(): void {
   metric.max(Metric.ClientTimestamp, ts);
   metric.max(Metric.Playback, config.lean ? BooleanFlag.False : BooleanFlag.True);
 
+  // Read cookies specified in configuration
```
toby-walker, line 35: What is this for? What cookies would be configured and why?

packages/clarity-js/src/data/metadata.ts
```diff
@@ -99,15 +109,19 @@ function num(string: string, base: number = 36): number {
 }
 
 function user(): number {
-  let id;
+  let id = cookie(Constant.STORAGE_KEY);
```
toby-walker, line 112: Shouldn't this be "USERID_KEY" and not STORAGE_KEY?

packages/clarity-js/types/data.d.ts
```diff
@@ -123,12 +124,14 @@ export const enum Constant {
...
     e: Token[];
-    d: Token[][];
+    a: Token[][];
+    p: Token[][];
 }
 
 export interface EncodedPayload {
     e: string;
-    d: string;
+    a: string;
```
toby-walker, line 133: Minor: These short forms are very cryptic. Add a brief comment annotating what goes where

packages/clarity-js/src/data/metadata.ts
```diff
@@ -99,15 +109,19 @@ function num(string: string, base: number = 36): number {
...
+}
+
+function cookie(key: string): string {
   let cookies: string[] = document.cookie.split(Constant.SEMICOLON);
   if (cookies) {
     for (let i = 0; i < cookies.length; i++) {
       let pair: string[] = cookies[i].split(Constant.EQUALS);
-      if (pair.length > 1 && pair[0].indexOf(Constant.STORAGE_KEY) >= 0 && pair[1].length === 8) {
-        id = pair[1];
+      if (pair.length > 1 && pair[0].indexOf(key) >= 0) {
```
toby-walker, line 121: Why match on just the partial string for key?

packages/clarity-js/src/data/upload.ts
```diff
@@ -39,35 +39,20 @@ export function queue(tokens: Token[], transmit: boolean = true): void {
...
-                }
-                break;
-            case Event.Upgrade:
-                // As part of upgrading experience from lean mode into full mode, we lookup anything that is backed up in memory
-                // from previous layout events and get them ready to go out to server as part of next upload.
-                for (let entry of backup) { container.push(entry); }
-                backup = [];
-                backupBytes = 0;
+                playbackBytes += event.length;
+                if (playbackBytes < MAX_PLAYBACK_BYTES) { playback.push(event); }
```
toby-walker, line 49: We should signal somehow that we have hit this limit and that the rest of the data is not complete.

packages/clarity-js/src/data/metadata.ts
```diff
@@ -20,6 +21,9 @@ export function start(): void {
     pageNum: s[1]
   }
 
+  // Check if the session should start off in full mode based on the signal from session storage
+  config.lean = config.track && sessionStorage && sessionStorage.getItem(Constant.UPGRADE_KEY) ? false : config.lean;
```
toby-walker, line 25: Why is the first case of the : false? If the condition is met should the lean mode not be turned to on?
Also is it safe to override the config as you do here?

lerna.json
sarveshnagpal: Our stable is 0.5.X and the current lean mode is 0.6.X and when it is generally available, it will be 0.6.0. All current iterations are beta.* per NPM conventions.

packages/clarity-decode/src/clarity.ts
sarveshnagpal: Because different providers (e.g. pointer vs. resize) can buffer events at a different rate and even though Clarity ensures timestamps are valid - i.e. when the event happened - they may not be able to go on wire at the exact time. We ensure that we relatively sort all the events before processing them further or visualizing them.

packages/clarity-js/src/data/metadata.ts
sarveshnagpal: Will forward an offline thread on this.

packages/clarity-js/src/data/metadata.ts
sarveshnagpal: Currently they are same constants. I will duplicate to avoid confusion.

packages/clarity-js/types/data.d.ts
sarveshnagpal: Ok sure, done.

packages/clarity-js/src/data/upload.ts
sarveshnagpal: This is now new code addition and the comment explains it is not ideal. Something we can tackle in future PRs.

packages/clarity-js/src/data/metadata.ts
sarveshnagpal: Fixed it to match full string.

packages/clarity-js/src/data/metadata.ts
sarveshnagpal: Because if the session was previously upgraded we want the lean mode to be off (false) for this session. If the session was not already upgraded then we honor whatever current config says. Yes, safe to do so.

packages/clarity-decode/src/clarity.ts
toby-walker: Please add the above in comment. 

---
In reply to: [467635660](https://github.com/microsoft/clarity/pull/31#discussion_r467635660) [](ancestors = 467635660)

packages/clarity-js/src/data/metadata.ts
toby-walker: No change made yet that I saw in iteration #6.

---
In reply to: [467635724](https://github.com/microsoft/clarity/pull/31#discussion_r467635724) [](ancestors = 467635724)

PR #33, opened 2020-08-27
Refactoring code and new features in 0.6.0-beta.6

sarveshnagpal:
- Refactoring global variable from clarity.* -> clarity(*) to queue up events on the client
- Enable unmasking through configuration
- Suppress selectors using config in addition to existing support for HTML attribute
- Simplify configuration and move internal settings to const enum
- Add event summary
- Add checks for limits and fail safe settings
- Log when we hit a limit and that causes Clarity to stop
- Add hash to each element during visualization
- Change metadata function as a callback


packages/clarity-js/types/core.d.ts
```diff
@@ -16,17 +16,25 @@ export const enum Time {
     Hour = 60 * 60 * 1000
 }
 
+export const enum Setting {
+    LongTask = 30, // 30ms
+}
```
swaathee, line 21: why not combine with the other setting enum?

packages/clarity-js/src/layout/dom.ts
```diff
@@ -42,18 +43,27 @@ function reset(): void {
...
+        for (const key of Object.keys(config.regions)) {
             let element = root.querySelector(config.regions[key]);
             if (element) { regionMap.set(element, key); }
         }
+
+        // Extract nodes with explicit masked configuration
+        for (const entry of config.suppress) {
+            let element = root.querySelector(entry);
+            if (element) { maskedMap.set(element, true); }
+        }
```
swaathee, line 66: what happens when mask is true and suppress is also set in config? : do we need to think about unmasking some elements only based on config?

packages/clarity-js/src/data/upload.ts
```diff
@@ -43,10 +42,11 @@ export function queue(tokens: Token[], transmit: boolean = true): void {
         switch (type) {
             case Event.Discover:
             case Event.Mutation:
-                // At the moment, we limit playback to grow until MAX_PLAYBACK_BYTES. Anytime we grow past this size, we start dropping events.
+                // At the moment, we limit playback to grow until MAX_PLAYBACK_BYTES.
+                // Anytime we grow past this size, we start dropping events.
                 // This is not ideal, and more of a fail safe mechanism.
```
swaathee, line 47: nit : change comment to reflect recent changes

packages/clarity-js/src/data/limit.ts
```diff
@@ -0,0 +1,38 @@
...
+        if (reason !== data.check) {
+            trigger(reason);
+        }
+    }
+}
+
+export function trigger(reason: Check): void {
+    data.check = reason;
+    clarity.stop();
+}
```
swaathee, line 28: how do we ensure compute is called before this?

sarveshnagpal: Discussed offline. Through module ordering in data/index.ts file.

packages/clarity-js/src/data/upload.ts
sarveshnagpal: Thanks, fixed.

PR #39, opened 2020-09-17
Change ids to string and selector bug fixes

sarveshnagpal:
- Change ids to be string instead of number
- Add selector field to Click event for debugging
- Capture text for non-link elements too
- Limit click text to 25 characters (configured by Setting)
- Remove backward compatibility for .next field while parsing DOM
- Fixing bug to compute selector correctly during decode logic
- Removing id field from computing selector

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -2,7 +2,7 @@
   "packages": [
     "packages/*"
   ],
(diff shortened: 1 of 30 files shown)


swaathee (approved):
:shipit:

PR #42, opened 2020-09-21
Changes to click event and privacy settings

sarveshnagpal:


packages/clarity-decode/src/interaction.ts
```diff
@@ -28,13 +28,14 @@ export function decode(tokens: Data.Token[]): InteractionEvent {
...
                 button: tokens[7] as number,
-                text: tokens[8] as string,
-                link: tokens[9] as string,
-                hash: tokens[10] as number,
-                region: tokens.length > 11 ? tokens[11] as number : null
+                count: tokens[8] as number,
+                context: tokens[9] as number,
+                text: tokens[10] as string,
+                link: tokens[11] as string,
+                element: tokens[12] as string,
```
narendrarana, line 35: Should we not add the same change element instead of hash to timeline event on the client?

PR #47, opened 2020-10-12
Masking and visualization updates

sarveshnagpal:
- Using ResizeObserver for computing size of parent of text nodes
- Placeholder image and IFRAME updates
- Support for compression when masking


packages/clarity-js/src/diagnostic/script.ts
```diff
@@ -21,6 +28,18 @@ function handler(error: ErrorEvent): void {
...
         };
+
+        // In certain cases, ResizeObserver could lead to flood of benign errors - especially when video element is involved.
+        // Reference Chromium issue: https://bugs.chromium.org/p/chromium/issues/detail?id=809574
+        // Even though it doesn't impact user experience, or show up in console, it can still flood error reporting through on error
+        // To mitigate that, we turn off Clarity's ResizeObserver on getting the first instance of this error
+        if (e.message.indexOf(Constant.ResizeObserver) >= 0) {
+            box.stop();
+            return false;
+        }
```
swaathee, line 39: does this also stop clarity and not just bm? (our computation of relative click positions will be impacted if we just stop bm)

swaathee (approved):
:shipit:

PR #54, opened 2020-10-26
Data quality bug fixes

sarveshnagpal:
- Override config with session data
- Do not attempt retries if server returns 400 bad payload


swaathee (approved):
:shipit:

PR #63, opened 2020-11-13
Fixing script error

sarveshnagpal:


packages/clarity-js/src/layout/dom.ts
```diff
@@ -339,9 +339,9 @@ export function getRegion(regionId: number): string {
 
 export function regions(): NodeValue[] {
     let v = [];
-    for (let id in values) {
-        if (values[id].metadata.active && values[id].metadata.region) {
-            v.push(values[id]);
+    for (let value of values) {
+        if (value && value.metadata.active && value.metadata.region) {
```
swaathee, line 343: >value.metadata [](start = 21, length = 14)

should we also null check metadata?

swaathee (approved):
:shipit:

packages/clarity-js/src/layout/dom.ts
sarveshnagpal: yes good point

PR #81, opened 2021-01-04
Adding visualization support for visibility signal

sarveshnagpal:
- Also, fixing script error due to upgrade call on pages where Clarity was inactive


toby-walker (approved):
:shipit:

PR #87, opened 2021-01-28
Simplifying regions, performance improvements & bug fixes

sarveshnagpal:
- Dropping bounding rectangle for regions
- Using IntersectionObserver
- Updating decode logic to match new data schema
- Stop listening to unload event to not impact backward-forward cache
- Get clicked text from buttons, input fields as well as images

### DIFF
--- a/packages/clarity-decode/src/layout.ts
+++ b/packages/clarity-decode/src/layout.ts
@@ -20,9 +20,11 @@ export function decode(tokens: Data.Token[]): LayoutEvent {
         case Data.Event.Region:
             let regionData: Layout.RegionData[] = [];
             for (let i = 2; i < tokens.length; i += 3) {
+                // For backward compatibility since 0.6.4, we extract visibility signal from the "box" field if it is present
+                let legacy = Array.isArray(tokens[i+1]) && (tokens[i+1] as number[]).length === 5;
                 let region: Layout.RegionData = {
                     id: tokens[i] as number,
-                    visible: tokens[i + 1] as number,
+                    visible: legacy ? tokens[i + 1][4] as number : tokens[i + 1] as number,
                     region: tokens[i + 2] as string
                 };
                 regionData.push(region);
--- a/packages/clarity-decode/src/layout.ts
+++ b/packages/clarity-decode/src/layout.ts
@@ -20,10 +20,9 @@ export function decode(tokens: Data.Token[]): LayoutEvent {
         case Data.Event.Region:
             let regionData: Layout.RegionData[] = [];
             for (let i = 2; i < tokens.length; i += 3) {
-                let box = tokens[i + 1] as number[];
                 let region: Layout.RegionData = {
                     id: tokens[i] as number,
-                    box: {x: box[0], y: box[1], w: box[2], h: box[3], v: box[4]},
+                    visible: tokens[i + 1] as number,
                     region: tokens[i + 2] as string
                 };
                 regionData.push(region);
--- a/packages/clarity-js/src/interaction/click.ts
+++ b/packages/clarity-js/src/interaction/click.ts
@@ -1,11 +1,11 @@
 import { BooleanFlag, Constant, Event, Setting } from "@clarity-types/data";
 import { BrowsingContext, ClickState } from "@clarity-types/interaction";
+import { Box } from "@clarity-types/layout";
 import { bind } from "@src/core/event";
 import { schedule } from "@src/core/task";
 import { time } from "@src/core/time";
 import { iframe } from "@src/layout/dom";
 import offset from "@src/layout/offset";
-import { layout } from "@src/layout/region";
 import { link, target } from "@src/layout/target";
 import encode from "./encode";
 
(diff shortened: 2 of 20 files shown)


packages/clarity-decode/src/layout.ts
```diff
@@ -20,10 +20,9 @@ export function decode(tokens: Data.Token[]): LayoutEvent {
         case Data.Event.Region:
             let regionData: Layout.RegionData[] = [];
             for (let i = 2; i < tokens.length; i += 3) {
-                let box = tokens[i + 1] as number[];
                 let region: Layout.RegionData = {
                     id: tokens[i] as number,
-                    box: {x: box[0], y: box[1], w: box[2], h: box[3], v: box[4]},
```
toby-walker, line 26: Wouldn't be backwards compatible. Any implications?

packages/clarity-js/src/interaction/click.ts
```diff
@@ -74,11 +74,15 @@ function handler(event: Event, root: Node, evt: MouseEvent): void {
...
 function text(element: Node): string {
     let output = null;
-    if (element && element.textContent) {
-        // Trim any spaces at the beginning or at the end of string
-        // Also, replace multiple occurrence of space characters with a single white space
-        // Finally, send only first few characters as specified by the Setting
-        output = element.textContent.trim().replace(/\s+/g, Constant.Space).substr(0, Setting.ClickText);
+    if (element) {
+        // Grab text using "textContent" for most HTMLElements, however, use "value" for HTMLInputElements.
+        let t = element.textContent || (element as HTMLInputElement).value;
```
toby-walker, line 79: Will this fail if it's not an HTMlInputElement?

packages/clarity-js/src/layout/encode.ts
```diff
@@ -26,12 +26,13 @@ export default async function (type: Event, ts: number = null): Promise<void> {
             queue(tokens);
             break;
         case Event.Region:
-            let bm = region.updates();
-            for (let value of bm) {
+            let r = region.data;
+            for (let value of r) {
                 tokens.push(value.id);
-                tokens.push([value.box.x, value.box.y, value.box.w, value.box.h, value.box.v]);
```
toby-walker, line 32: Back/forwards compat?

packages/clarity-js/src/layout/index.ts
```diff
@@ -7,17 +7,17 @@ import * as region from "@src/layout/region";
...
+    region.start();
     dom.start();
     mutation.start();
     discover.start();
     box.start();
-    region.reset();
 }
 
 export function stop(): void {
+    region.stop();
```
toby-walker, line 18: Why is order of these in start/stop different and is that intentional? If intentional, could you add a quick comment why?

packages/clarity-js/src/layout/region.ts
```diff
@@ -1,108 +1,77 @@
...
-        update(value.id, layout(dom.getNode(value.id) as Element, viewport));
-    }
-
-    if (updateMap.length > 0) { await encode(Event.Region); }
+export function start(): void {
+    reset();
+    observer = null;
+    regionMap = new WeakMap();
+    regions = {};
+    watch = window["IntersectionObserver"] ? true : false;
```
toby-walker, line 18: Will this lower region coverage based on intersection observer support in browsers?

packages/clarity-js/src/layout/region.ts
```diff
@@ -1,108 +1,77 @@
...
-                w: Math.floor(rect.width),
-                h: Math.floor(rect.height),
-                v: visible ? BooleanFlag.True : BooleanFlag.False
-            };
+function handler(entries: IntersectionObserverEntry[]): void {
+    for (let entry of entries) {
+        let target = entry.target;
+        let id = target ? dom.getId(target) : null;
+        if (id && regionMap.has(target)) {
+            let visible = entry.isIntersecting ? BooleanFlag.True : BooleanFlag.False;
```
toby-walker, line 54: Have you thought about sending interactoinRatio and/or interactionRect too? You probably did but would be interested in the thinking.

packages/clarity-js/src/layout/region.ts
```diff
@@ -1,108 +1,77 @@
...
+    
 }
 
-export function updates(): RegionData[] {
-    let summary = [];
-    for (let id of updateMap) {
-        summary.push(bm[id]);
+export function observe(node: Node, name: string): void {
+    if (regionMap.has(node) === false) {
+        regionMap.set(node, name);
```
toby-walker, line 24: This will use the default threshold of zero for the thresholds. Will that causes a lot of events and would a higher threshold be appropriate? By "lot of events" I mean on the client and also events we upload.

toby-walker (changes requested):
:clock1:

packages/clarity-decode/src/layout.ts
sarveshnagpal: Good point; will make it backward compatible.

packages/clarity-js/src/interaction/click.ts
sarveshnagpal: It wouldn't since the value will come back as undefined and it won't pass the next line.

packages/clarity-js/src/layout/encode.ts
sarveshnagpal: I will tackle this as part of decode.

packages/clarity-js/src/layout/index.ts
sarveshnagpal: It is based on the dependency order. I will add comment.

packages/clarity-js/src/layout/region.ts
sarveshnagpal: No real impact since this API has been around for a while: https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API#browser_compatibility

packages/clarity-js/src/layout/region.ts
sarveshnagpal: Yes but intentionally favoring simplicity here to keep script size small. We can always extend when we have a use case.

packages/clarity-js/src/layout/region.ts
sarveshnagpal: Zero is intentional because it's hard to come up with a magic number otherwise. We tweak this later but this worked best when dealing with regions of various sizes. For instance 10% of "body" may not trigger visibility if the page was long and that will be inaccurate to say that body wasn't visible. Also, it won't result in more events. If threshold is zero and once it becomes visible, next event only fires when it gets hidden. Basically no chatter as it goes from 0% -> 10% -> 50% -> 100%.

packages/clarity-decode/src/layout.ts
sarveshnagpal: Done

swaathee (approved):
:shipit:

toby-walker (approved):
:shipit:

PR #103, opened 2021-02-12
Visualization updates and regions bug fixes

sarveshnagpal:
- On hover visualization fix
- Rendering cross-origin iframe as blocked
- Accepting multiple regions per page
- Bug fixing race condition between region processing and DOM discovery

### DIFF
--- a/packages/clarity-js/src/layout/dom.ts
+++ b/packages/clarity-js/src/layout/dom.ts
@@ -1,8 +1,9 @@
 import { Privacy } from "@clarity-types/core";
-import { Setting } from "@clarity-types/data";
+import { Code, Setting, Severity } from "@clarity-types/data";
 import { Constant, NodeChange, NodeInfo, NodeValue, Source } from "@clarity-types/layout";
 import config from "@src/core/config";
 import { time } from "@src/core/time";
+import * as log from "@src/diagnostic/log";
 import * as region from "@src/layout/region";
 import selector from "@src/layout/selector";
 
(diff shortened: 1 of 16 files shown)


packages/clarity-js/src/layout/dom.ts
```diff
@@ -54,11 +54,13 @@ function reset(): void {
 export function parse(root: ParentNode): void {
     // Since mutations may happen on leaf nodes too, e.g. text nodes, which may not support all selector APIs.
     // We ensure that the root note supports querySelector API before executing the code below to identify new regions.
```
toby-walker, line 56: >querySelector [](start = 45, length = 13)

Nit: queyrSelectorAll

toby-walker:
> }

Missing version bump?

---
Refers to: packages/clarity-js/src/interaction/scroll.ts:58 in 8c24e46. [](commit_id = 8c24e467e84fd57b24e7cb058e2dfa68b613862d, deletion_comment = False)

packages/clarity-js/src/layout/region.ts
```diff
@@ -42,23 +44,39 @@ export function track(id: number): void {
...
 
 export function compute(): void {
+    // Process any regions that couldn't be processed earlier
+    let q = [];
+    for (let r of queue) {
+        let id = dom.getId(r.node);
+        if (id) {
+            r.data.id = id;
+            data.push(r.data);
+        } else { q.push(r); }
```
toby-walker, line 54: This looks like it could get stuck in an ever growing queue if things did not work out. Perhaps limit the number of retries or  timeout them out.

packages/clarity-js/src/layout/dom.ts
sarveshnagpal: Thanks, updated.

packages/clarity-js/src/layout/region.ts
```diff
@@ -42,23 +44,39 @@ export function track(id: number): void {
...
 
 function handler(entries: IntersectionObserverEntry[]): void {
     for (let entry of entries) {
         let target = entry.target;
-        let id = target ? dom.getId(target) : null;
-        if (id && regionMap.has(target)) {
+        let rect = entry.boundingClientRect;
+
+        // Only capture regions that are not hidden on the page
+        if (regionMap.has(target) && rect.width > 0 && rect.height > 0) {
```
toby-walker, line 68: What if the region is off document? Off screen? Looked around to see how this was done and saw lot of methods, many of which are much more complicated than this. Some links:

[https://stackoverflow.com/questions/66070706/how-to-determine-if-html-element-is-generally-visible-javascript](https://stackoverflow.com/questions/66070706/how-to-determine-if-html-element-is-generally-visible-javascript) (lot of complicated logic)
[https://www.javascripttutorial.net/dom/css/check-if-an-element-is-visible-in-the-viewport/](https://www.javascripttutorial.net/dom/css/check-if-an-element-is-visible-in-the-viewport/) (use inner wdith/heigh, etc.)
[https://www.geeksforgeeks.org/how-to-check-if-an-element-is-visible-in-dom/](https://www.geeksforgeeks.org/how-to-check-if-an-element-is-visible-in-dom/) (sort of like your but more)

Why does interaction observer even return hidden elements? Seems odd. Also would any of the intersection observe entry properties be useful for this check (they already exist, no bounding box computation). 

Etc.

toby-walker (changes requested):
:clock1:

packages/clarity-js/src/layout/region.ts
sarveshnagpal: We are using entry's bounding rectangle for the check - not recomputing. The goal here is to eliminate low hanging fruit - for instance, elements that are on the page but are display:none. Edge cases are both a significant hit to performance (computedStyles) and also not guaranteed. Plus we also don't need to cover all edge cases since these rules are for user defined regions and can be modified to tackle edge cases.

packages/clarity-js/src/layout/region.ts
sarveshnagpal: Discussed this offline and we are in sync. This will be capped at number of the regions and I've added more comments to reflect the intent of this code path.

toby-walker (approved):
:shipit:

swaathee (approved):
:shipit:

PR #111, opened 2021-02-18
Handle case when user clicks an input that has data in it

harip:
When a user clicks an existing input box and clicks out of it, even with mask tag set to true, data was being sent as unmasked.


sarveshnagpal (approved):

sarveshnagpal:
Thanks for finding and fixing this @harip.

PR #112, opened 2021-02-19
additional metadata per page

ender336:


packages/clarity-js/src/data/metadata.ts
```diff
@@ -32,6 +32,11 @@ export function start(): void {
   dimension.log(Dimension.PageTitle, title);
   dimension.log(Dimension.Url, location.href);
   dimension.log(Dimension.Referrer, document.referrer);
+  dimension.log(Dimension.Language, (<any>navigator).userLanguage || navigator.language);
+  dimension.log(Dimension.ScreenWidth, `${screen.width || Constant.Empty}`);
```
swaathee, line 36: is this different from the viewport width and height that we already capture?

swaathee (changes requested):

packages/clarity-js/src/data/metadata.ts
```diff
@@ -139,3 +144,47 @@ function cookie(key: string): string {
   }
   return null;
 }
+
+function addPerformanceDimensions(): void {
```
sarveshnagpal, line 148: We already send this as part of performance module. We should use that data, instead of sending again.  See: https://github.com/microsoft/clarity/blob/master/packages/clarity-js/src/performance/navigation.ts

packages/clarity-js/types/data.d.ts
```diff
@@ -75,7 +75,14 @@ export const enum Dimension {
     SchemaType = 5,
     ProductBrand = 6,
     ProductAvailability = 7,
-    AuthorName = 8
+    AuthorName = 8,
+    Language = 9,
+    ScreenWidth = 10,
```
sarveshnagpal, line 80: These should be metric; not dimension. Dimension => string & Metric => number.

sarveshnagpal (changes requested):

packages/clarity-js/src/data/metadata.ts
ender336: yes actually, they are constant for a monitor rather than specific to a window's size

packages/clarity-js/src/data/metadata.ts
```diff
@@ -32,10 +32,14 @@ export function start(): void {
   dimension.log(Dimension.PageTitle, title);
   dimension.log(Dimension.Url, location.href);
   dimension.log(Dimension.Referrer, document.referrer);
+  dimension.log(Dimension.Language, (<any>navigator).userLanguage || navigator.language);
```
sarveshnagpal, line 35: Please add null check to prevent against unknowns.

sarveshnagpal (approved):
Please feel free to merge after adding null checks for navigation & screen. Thanks.

swaathee (approved):

packages/clarity-js/src/data/metadata.ts
```diff
@@ -32,6 +32,11 @@ export function start(): void {
...
+  dimension.log(Dimension.ScreenWidth, `${screen.width || Constant.Empty}`);
+  dimension.log(Dimension.ScreenHeight, `${screen.height || Constant.Empty}`);
```
toby-walker: Why are we logging an integer as a string/

packages/clarity-js/types/data.d.ts
toby-walker: +1

---
In reply to: [579243873](https://github.com/microsoft/clarity/pull/112#discussion_r579243873) [](ancestors = 579243873)

PR #114, opened 2021-02-23
Refactoring code to parse JSON+LD

sarveshnagpal:
- Adding new fields
- Refactoring JSON+LD code to a separate module
- Data quality bug fixes

### DIFF
--- a/packages/clarity-js/src/layout/schema.ts
+++ b/packages/clarity-js/src/layout/schema.ts
@@ -1,0 +1,56 @@
+import { Dimension, Metric, Setting } from "@clarity-types/data";
+import { Constant, JsonLD } from "@clarity-types/layout";
+import * as dimension from "@src/data/dimension";
+import * as metric from "@src/data/metric";
+
+/* JSON+LD (Linked Data) Recursive Parser */
+export function ld(json: any): void {
+    for (let key of Object.keys(json)) {
+        let value = json[key];
+        if (key === JsonLD.Type && typeof value === "string") {
+            value = value.toLowerCase();
+            /* Normalizations */
+            value = value.indexOf(JsonLD.Article) >= 0 || value.indexOf(JsonLD.Posting) >= 0 ? JsonLD.Article : value;
+            switch (value) {
+                case JsonLD.Article:
+                case JsonLD.Recipe:
+                    dimension.log(Dimension.SchemaType, json[key]);
+                    break;
+                case JsonLD.Product:
+                    dimension.log(Dimension.SchemaType, json[key]);
+                    dimension.log(Dimension.ProductName, json[JsonLD.Name]);
+                    dimension.log(Dimension.ProductSku, json[JsonLD.Sku]);
+                    if (json[JsonLD.Brand]) { dimension.log(Dimension.ProductBrand, json[JsonLD.Brand][JsonLD.Name]); }
+                    break;
+                case JsonLD.Rating:
+                    if (json[JsonLD.RatingValue]) {
+                        let bestRating = num(json[JsonLD.BestRating], Setting.BestRatingValue);
+                        let rating = (parseFloat(json[JsonLD.RatingValue]) * Setting.RatingScale) / bestRating;
+                        metric.max(Metric.RatingValue, rating);
+                    }
+                    metric.max(Metric.RatingCount, num(json[JsonLD.RatingCount]));
+                    metric.max(Metric.ReviewCount, num(json[JsonLD.ReviewCount]));
+                    break;
+                case JsonLD.Author:
+                    dimension.log(Dimension.AuthorName, json[JsonLD.Name]);
+                    break;
+                case JsonLD.Offer:
+                    dimension.log(Dimension.ProductAvailability, json[JsonLD.Availability]);
+                    dimension.log(Dimension.ProductCondition, json[JsonLD.ItemCondition]);
+                    dimension.log(Dimension.Currency, json[JsonLD.PriceCurrency]);
+                    dimension.log(Dimension.ProductSku, json[JsonLD.Sku]);
+                    metric.max(Metric.ProductPrice, num(json[JsonLD.Price]));
+                    break;
+                case JsonLD.Brand:
+                    dimension.log(Dimension.ProductBrand, json[JsonLD.Name]);
+                    break;
+            }
+        }
+        // Continue parsing nested objects
+        if (value !== null && typeof(value) === Constant.Object) { ld(value); }
+    }
+}
+
+function num(input: string, backup: number = null): number {
+    return input ? Math.round(parseFloat(input)) : backup;
+}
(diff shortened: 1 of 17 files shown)


packages/clarity-js/types/layout.d.ts
```diff
@@ -68,10 +68,18 @@ export const enum JsonLD {
     Offer = "offer",
     Brand = "brand",
     RatingValue = "ratingValue",
+    BestRating = "bestRating",
```
toby-walker, line 71: worstRating also needed.

packages/clarity-js/src/layout/schema.ts
```diff
@@ -0,0 +1,56 @@
...
+                    metric.max(Metric.RatingCount, num(json[JsonLD.RatingCount]));
+                    metric.max(Metric.ReviewCount, num(json[JsonLD.ReviewCount]));
+                    break;
+                case JsonLD.Author:
+                    dimension.log(Dimension.AuthorName, json[JsonLD.Name]);
+                    break;
+                case JsonLD.Offer:
+                    dimension.log(Dimension.ProductAvailability, json[JsonLD.Availability]);
+                    dimension.log(Dimension.ProductCondition, json[JsonLD.ItemCondition]);
+                    dimension.log(Dimension.Currency, json[JsonLD.PriceCurrency]);
```
toby-walker, line 40: Very, very Minor but could we call this PriceCurrency to keep naming consistent.

swaathee (approved):
:shipit:

packages/clarity-js/src/layout/schema.ts
```diff
@@ -0,0 +1,56 @@
...
+                    break;
+                case JsonLD.Product:
+                    dimension.log(Dimension.SchemaType, json[key]);
+                    dimension.log(Dimension.ProductName, json[JsonLD.Name]);
+                    dimension.log(Dimension.ProductSku, json[JsonLD.Sku]);
+                    if (json[JsonLD.Brand]) { dimension.log(Dimension.ProductBrand, json[JsonLD.Brand][JsonLD.Name]); }
+                    break;
+                case JsonLD.Rating:
+                    if (json[JsonLD.RatingValue]) {
+                        let bestRating = num(json[JsonLD.BestRating], Setting.BestRatingValue);
```
toby-walker, line 27: >BestRating [](start = 57, length = 10)

Needs to use the worst rating as well. Where it default to 1, best to 5 and then you do (rating-worstRating)/(bestRating-WorstRating)*5 + 1.

packages/clarity-js/src/layout/schema.ts
```diff
@@ -0,0 +1,56 @@
...
+                        let rating = (parseFloat(json[JsonLD.RatingValue]) * Setting.RatingScale) / bestRating;
```
toby-walker, line 28: >bestRating [](start = 100, length = 10)

Needs to be +1 here assuming worstRAting default is 1 and the scale is [1,5]

packages/clarity-js/src/layout/schema.ts
```diff
@@ -0,0 +1,56 @@
...
+import * as metric from "@src/data/metric";
+
+/* JSON+LD (Linked Data) Recursive Parser */
+export function ld(json: any): void {
+    for (let key of Object.keys(json)) {
+        let value = json[key];
+        if (key === JsonLD.Type && typeof value === "string") {
+            value = value.toLowerCase();
+            /* Normalizations */
+            value = value.indexOf(JsonLD.Article) >= 0 || value.indexOf(JsonLD.Posting) >= 0 ? JsonLD.Article : value;
```
toby-walker, line 13: Can you just do === as this indexOf seems like it might over-match.

swaathee: I think that the value will have the article / post id as well which is why we need to do the normalization 

---
In reply to: [581303085](https://github.com/microsoft/clarity/pull/114#discussion_r581303085) [](ancestors = 581303085)

packages/clarity-js/src/layout/schema.ts
```diff
@@ -0,0 +1,56 @@
...
+                case JsonLD.Recipe:
+                    dimension.log(Dimension.SchemaType, json[key]);
+                    break;
+                case JsonLD.Product:
+                    dimension.log(Dimension.SchemaType, json[key]);
+                    dimension.log(Dimension.ProductName, json[JsonLD.Name]);
+                    dimension.log(Dimension.ProductSku, json[JsonLD.Sku]);
+                    if (json[JsonLD.Brand]) { dimension.log(Dimension.ProductBrand, json[JsonLD.Brand][JsonLD.Name]); }
+                    break;
+                case JsonLD.Rating:
```
toby-walker, line 25: >Rating [](start = 28, length = 6)

Is this AggregateRatingg? If so please rename to full type name.

packages/clarity-js/src/layout/schema.ts
toby-walker, line 13: Recommend values to normalize to article are:

* NewsArticle
* BlobPosting

See [https://schema.org/docs/full.html](https://schema.org/docs/full.html)

If it's just the constant names that differ can you please update them to the full name so it's easier to match?

toby-walker (changes requested):
:clock1:

packages/clarity-js/src/layout/schema.ts
```diff
@@ -0,0 +1,56 @@
...
+                    dimension.log(Dimension.AuthorName, json[JsonLD.Name]);
+                    break;
+                case JsonLD.Offer:
+                    dimension.log(Dimension.ProductAvailability, json[JsonLD.Availability]);
+                    dimension.log(Dimension.ProductCondition, json[JsonLD.ItemCondition]);
+                    dimension.log(Dimension.Currency, json[JsonLD.PriceCurrency]);
+                    dimension.log(Dimension.ProductSku, json[JsonLD.Sku]);
+                    metric.max(Metric.ProductPrice, num(json[JsonLD.Price]));
+                    break;
+                case JsonLD.Brand:
```
toby-walker, line 44: These rely on typing only but not the "path" to the property. Example: brand.name is the property path. If the type of the brand property is not "Brand" then we wouldn't pick up the attribute. This would complicate this a bit but it would be ideal if the code kept the "current" path around and let us use that for the matching as well as the type name. 

Ignore if you think this is too much.

packages/clarity-js/src/layout/schema.ts
sarveshnagpal: Yes sure

packages/clarity-js/types/layout.d.ts
sarveshnagpal: Good point, done.

packages/clarity-js/src/layout/schema.ts
sarveshnagpal: Done

packages/clarity-js/src/layout/schema.ts
sarveshnagpal: Yeah let's come back to it later.

toby-walker (approved):
:shipit:

PR #118, opened 2021-03-03
Improve region handling and bug fixes

sarveshnagpal:
- Better intersection inference
- Added support for region visualization in the extension
- Fix script error with performance observer in older browsers

### DIFF
--- a/packages/clarity-js/src/layout/region.ts
+++ b/packages/clarity-js/src/layout/region.ts
@@ -1,4 +1,4 @@
-import { BooleanFlag, Event } from "@clarity-types/data";
+import { BooleanFlag, Event, Setting } from "@clarity-types/data";
 import { RegionData, RegionQueue } from "@clarity-types/layout";
 import * as dom from "@src/layout/dom";
 import encode from "@src/layout/encode";
@@ -24,7 +24,11 @@ export function start(): void {
 export function observe(node: Node, name: string): void {
     if (regionMap.has(node) === false) {
         regionMap.set(node, name);
-        observer = observer === null && watch ? new IntersectionObserver(handler) : observer;
+        observer = observer === null && watch ? new IntersectionObserver(handler, {
+            // Get notified as intersection continues to change
+            // This allows us to process regions that get partially hidden during the lifetime of the page
+            threshold: [0,0.2,0.4,0.6,0.8,1]
+        }) : observer;
         if (observer && node && node.nodeType === Node.ELEMENT_NODE) {
             observer.observe(node as Element);
         }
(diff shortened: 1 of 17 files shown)


packages/clarity-js/src/performance/observer.ts
```diff
@@ -23,16 +23,24 @@ export function start(): void {
...
+        // This is why we need to individually "observe" each supported type
+        for (let x of types) {
+            if (PerformanceObserver.supportedEntryTypes.indexOf(x) >= 0) {
+                // Initialize CLS with a value of zero. It's possible (and recommended) for sites to not have any cumulative layout shift.
+                // In those cases, we want to still initialize the metric in Clarity
+                if (x === Constant.CLS) { metric.sum(Metric.CumulativeLayoutShift, 0); }
+                observer.observe({type: x, buffered: true});
+            }
+        }
+    } catch { log.log(Code.PerformanceObserver, null, Severity.Warning); }
```
toby-walker, line 43: >null [](start = 48, length = 4)

Add the type here?

packages/clarity-js/src/layout/region.ts
```diff
@@ -24,7 +24,11 @@ export function start(): void {
 export function observe(node: Node, name: string): void {
     if (regionMap.has(node) === false) {
         regionMap.set(node, name);
-        observer = observer === null && watch ? new IntersectionObserver(handler) : observer;
+        observer = observer === null && watch ? new IntersectionObserver(handler, {
+            // Get notified as intersection continues to change
+            // This allows us to process regions that get partially hidden during the lifetime of the page
```
toby-walker, line 29: The defrault is 0 which means we should get called every time. Why is this better? 

From: [https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API)

Either a single number or an array of numbers which indicate at what percentage of the target's visibility the observer's callback should be executed. If you only want to detect when visibility passes the 50% mark, you can use a value of 0.5. If you want the callback to run every time visibility passes another 25%, you would specify the array [0, 0.25, 0.5, 0.75, 1]. The default is 0 (meaning as soon as even one pixel is visible, the callback will be run). A value of 1.0 means that the threshold isn't considered passed until every pixel is visible.

packages/clarity-js/src/layout/region.ts
```diff
@@ -72,20 +76,25 @@ function handler(entries: IntersectionObserverEntry[]): void {
...
+        if (regionMap.has(target) && rect.width > 0 && rect.height > 0 && viewport.width > 0 && viewport.height > 0) {
             let id = target ? dom.getId(target) : null;
-            let visible = entry.isIntersecting ? BooleanFlag.True : BooleanFlag.False;
-            let d = { id, visible, region: regionMap.get(target) };
-            if (id) {
-                regions[id] = d;
-                updates.push(id);
-                data.push(d);
-            } else { queue.push({node: target, data: d}); }
+            let area = overlap ? (overlap.width * overlap.height * 1.0) / (viewport.width * viewport.height) : 0;
```
toby-walker, line 87: >width [](start = 42, length = 5)

This uses overlap of viewport. The API ([https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserverEntry](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserverEntry)) provide insectionRatio, which is the ratio of intersectRect to boungingRect. Why prefer viewport ratio instead?

toby-walker, line 87: >area [](start = 16, length = 4)

Minor: Rename to viewportRatio or similar as it's not an area.

packages/clarity-js/src/layout/region.ts
```diff
@@ -72,20 +76,25 @@ function handler(entries: IntersectionObserverEntry[]): void {
...
+            let visible = area > Setting.IntersectionArea || entry.intersectionRatio > Setting.IntersectionRatio ? BooleanFlag.True : BooleanFlag.False;
```
toby-walker, line 88: >IntersectionArea [](start = 41, length = 16)

Rename to something clearer that this is for viewport, not general interaction.

toby-walker (approved):
:shipit:

packages/clarity-js/src/layout/region.ts
sarveshnagpal: https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API#creating_an_intersection_observer

I will clarify this in comments. Thanks.

packages/clarity-js/src/layout/region.ts
sarveshnagpal: Good point - done. Also explained why viewport ration - mainly to differentiate larger regions from smaller regions.

sarveshnagpal: Done

packages/clarity-js/src/layout/region.ts
sarveshnagpal: Done

packages/clarity-js/src/performance/observer.ts
sarveshnagpal: This is a call to the function that defines the type for second argument (Error) which is null here.

---
In reply to: [586552177](https://github.com/microsoft/clarity/pull/118#discussion_r586552177) [](ancestors = 586552177)

swaathee (approved):
:shipit:

PR #121, opened 2021-03-16
Support for metrics via DOM

sarveshnagpal:


packages/clarity-devtools/src/classification.ts
```diff
@@ -0,0 +1,13 @@
+import { Core } from "clarity-js";
+
+export default function(): Core.Config {
```
toby-walker, line 3: What does this represent?

packages/clarity-js/src/data/metric.ts
```diff
@@ -1,8 +1,10 @@
-import { Event, Metric, MetricData } from "@clarity-types/data";
+import { Code, Constant, Event, Metric, MetricData, Severity } from "@clarity-types/data";
+import * as log from "@src/diagnostic/log";
 import encode from "./encode";
 
 export let data: MetricData = null;
 export let updates: MetricData = null;
+const regex = /[^0-9\.]/g;
```
toby-walker, line 7: Minor: Doesn't' work with non-US numbers. Suggest you allow comma (,) as well.

toby-walker:
> export function parse(root: ParentNode): void {

Not specific to this change but depending on the mutation pattern this might miss things if the selector matches only after a subtree is available later even if the root node is available now.

---
Refers to: packages/clarity-js/src/layout/dom.ts:56 in d10b6bf. [](commit_id = d10b6bfb9484d9b56f1bc36370f92aa08268adc5, deletion_comment = False)

packages/clarity-js/types/core.d.ts
```diff
@@ -51,6 +51,11 @@ export interface Regions {
     [key: string]: string;
 }
 
+export interface Metrics {
+    /* In the following key-value pair, key is the given CSS Selector and value is Metric enum */
+    [key: string]: number;
```
toby-walker, line 56: >string [](start = 10, length = 6)

Shouldn't this be a number as it's used that way (I think).

packages/clarity-js/types/data.d.ts
```diff
@@ -69,7 +69,8 @@ export const enum Metric {
     ColorDepth = 16,
     ReviewCount = 17,
     BestRating = 18,
-    WorstRating = 19
+    WorstRating = 19,
+    CartTotal = 20
```
toby-walker, line 73: Since these are driven by configuration should they not be declared here? Can we, instead, reserve a range of value here that configuration is limited to?

toby-walker (changes requested):
:clock1:

packages/clarity-js/types/core.d.ts
sarveshnagpal: The comment is valid. In Javascript we can't have keys as numbers; they can only be strings. So I have selector as a key and metric as a value, so we can keep it as number. 

---
In reply to: [595644577](https://github.com/microsoft/clarity/pull/121#discussion_r595644577) [](ancestors = 595644577)

packages/clarity-js/types/data.d.ts
sarveshnagpal: We can technically override any metric value. For instance even ProductPrice or others similar to regions. We will add more checks as we get better understanding of it. 

---
In reply to: [595644826](https://github.com/microsoft/clarity/pull/121#discussion_r595644826) [](ancestors = 595644826)

sarveshnagpal:
> export function parse(root: ParentNode): void {

Great point on the edge case. I will brainstorm more on this offline.

---
In reply to: [800718489](https://github.com/microsoft/clarity/pull/121#issuecomment-800718489) [](ancestors = 800718489)

---
Refers to: packages/clarity-js/src/layout/dom.ts:56 in d10b6bf. [](commit_id = d10b6bfb9484d9b56f1bc36370f92aa08268adc5, deletion_comment = False)

packages/clarity-js/src/data/metric.ts
sarveshnagpal: Fixed thanks.

toby-walker (approved):
:shipit:

PR #122, opened 2021-03-18
client changes for visualize and bug fix for document width

swaathee:
Visualize changes : 
1. Render html at a give time 
2. Get time of the current snapshot

Bug fix
1. Document width is 0 for some sites

### DIFF
--- a/packages/clarity-js/src/layout/document.ts
+++ b/packages/clarity-js/src/layout/document.ts
@@ -16,7 +16,15 @@ export function start(): void {
 export function compute(): void {
     let body = document.body;
     let d = document.documentElement;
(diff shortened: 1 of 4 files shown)


sarveshnagpal:
> export function render(events: Data.DecodedEvent[]): void {

For consistency I think we should also update renderTime in this function. Not a blocker for now though.

---
Refers to: packages/clarity-visualize/src/clarity.ts:87 in b5d45d0. [](commit_id = b5d45d01b7357becbfed1de5e4c1bdc8e15db9b9, deletion_comment = False)

sarveshnagpal (approved):

PR #123, opened 2021-03-22
Display console log warning if website has multiple clarity scripts injected

saghoneim:


packages/clarity-js/src/global.ts
```diff
@@ -6,6 +6,7 @@ import * as clarity from "@src/clarity";
 (function(): void {
     if (typeof window !== "undefined") {
         const w = window as any;
+        if (!w.clarity.q) { console.log("Error CL001: Multiple Clarity tags detected."); }
```
swaathee, line 9: should we stop activating the second tag when the first is present?

saghoneim: I don't see a value in doing so, since in this case, the order in which the scripts run is not consistent. For simplicity, we could just leave it run and display the error code so the user is notified to fix this bad state.

sarveshnagpal: Thanks for the change. This will throw a script error if someone included clarity-js browser bundle without the initial tag. I think we should also check for w.clarity before looking up w.clarity.q. Also, I think it's better to not activate two bundles on the same page. Since we are showing a console error already - it will help explain the site owner that something with the setup is not right. Probably better for us to brainstorm offline before merging this change.

saghoneim: I agree on checking the w.clarity first to make sure that it's triggered from our script. If it doesn't exist we shouldn't throw this error. I'll ping you offline to discuss the 2 bundles being active together issue.

PR #127, opened 2021-04-09
Changes to sessionization logic and price extraction

sarveshnagpal:
- Moving to session cookie instead of session storage since it's unreliable on iOS Safari
- Bumping up the version to 0.6.11
- Adding support for TabId as part of a dimension (still through Session Storage)
- Updating cookie handling to be forward compatible for future changes

### DIFF
--- a/packages/clarity-js/src/data/metadata.ts
+++ b/packages/clarity-js/src/data/metadata.ts
@@ -19,19 +19,29 @@ export function start(): void {
   data = {
     projectId: config.projectId || hash(location.host),
     userId: user(),
-    sessionId: s.id,
+    sessionId: s.session,
     pageNum: s.count
+  }
+
+  // For backward compatibility; remove in future iterations (v0.6.11)
+  if (config.upload && typeof config.upload === Constant.String && (config.upload as string).indexOf(Constant.HTTPS) === 0) {
+    let url = config.upload as string;
+    config.server = url.substr(0, url.indexOf("/", Constant.HTTPS.length));
+    config.upload = config.server.length > 0 && config.server.length < url.length ? url.substr(config.server.length + 1) : url;
   }
 
   // Override configuration based on what's in the session storage
   config.lean = config.track && s.upgrade === BooleanFlag.True ? false : config.lean;
   config.upload = config.track && typeof config.upload === Constant.String && s.upload ? s.upload : config.upload;
 
+
   // Log dimensions
   dimension.log(Dimension.UserAgent, ua);
   dimension.log(Dimension.PageTitle, title);
   dimension.log(Dimension.Url, location.href);
   dimension.log(Dimension.Referrer, document.referrer);
+  dimension.log(Dimension.TabId, tab());
+  dimension.log(Dimension.PageLanguage, document.documentElement.lang);
   if (navigator) {
     dimension.log(Dimension.Language, (<any>navigator).userLanguage || navigator.language);
   }
@@ -47,7 +57,7 @@ export function start(): void {
 
   // Read cookies specified in configuration
   for (let key of config.cookies) {
-    let value = cookie(key);
+    let value = getCookie(key);
     if (value) { set(key, value); }
   }
 
@@ -71,8 +81,18 @@ export function consent(): void {
 }
 
 export function clear(): void {
-  // Clear any stored information in the session so we can restart fresh the next time
-  if (supported(window, Constant.SessionStorage)) { sessionStorage.removeItem(Constant.StorageKey); }
+  // Clear any stored information in the session cookie so we can restart fresh the next time
+  setCookie(Constant.SessionKey, Constant.Empty, 0);
+}
+
+function tab(): string {
+  let id = shortid();
+  if (supported(window, Constant.SessionStorage)) {
+    let value = sessionStorage.getItem(Constant.TabKey);
+    id = value && value.indexOf(Constant.Separator) < 0 ? value : id;
+    sessionStorage.setItem(Constant.TabKey, id);
+  }
+  return id;
 }
 
 export function save(): void {
@@ -80,8 +100,8 @@ export function save(): void {
   let upgrade = config.lean ? BooleanFlag.False : BooleanFlag.True;
   let upload = typeof config.upload === Constant.String ? config.upload : Constant.Empty;
   if (upgrade && callback) { callback(data, !config.lean); }
-  if (config.track && supported(window, Constant.SessionStorage)) {
-    sessionStorage.setItem(Constant.StorageKey, [data.sessionId, ts, data.pageNum, upgrade, upload].join(Constant.Separator));
+  if (config.track && supported(document, Constant.Cookie)) {
+    setCookie(Constant.SessionKey, [data.sessionId, ts, data.pageNum, upgrade, upload].join(Constant.Separator), Setting.SessionExpire);
   }
 }
 
@@ -90,13 +110,7 @@ function supported(target: Window | Document, api: string): boolean {
 }
 
 function track(): void {
-  if (config.track && supported(document, Constant.Cookie)) {
-    let expiry = new Date();
-    expiry.setDate(expiry.getDate() + Setting.Expire);
-    let expires = expiry ? Constant.Expires + expiry.toUTCString() : Constant.Empty;
-    let value = `${data.userId}${Constant.Semicolon}${expires}${Constant.Path}`;
-    document.cookie = Constant.CookieKey + Constant.Equals + value;
-  }
+  setCookie(Constant.CookieKey, data.userId, Setting.Expire);
 }
 
 function shortid(): string {
@@ -108,17 +122,16 @@ function shortid(): string {
 }
 
 function session(): Session {
-  let output: Session = { id: shortid(), ts: Math.round(Date.now()), count: 1, upgrade: BooleanFlag.False, upload: Constant.Empty };
-  if (config.track && supported(window, Constant.SessionStorage)) {
-    let value = sessionStorage.getItem(Constant.StorageKey);
-    if (value && value.indexOf(Constant.Separator) >= 0) {
-      let parts = value.split(Constant.Separator);
-      if (parts.length === 5 && output.ts - num(parts[1]) < Setting.SessionTimeout) {
-        output.id = parts[0];
-        output.count = num(parts[2]) + 1;
-        output.upgrade = num(parts[3]);
-        output.upload = parts[4];
-      }
+  let output: Session = { session: shortid(), ts: Math.round(Date.now()), count: 1, upgrade: BooleanFlag.False, upload: Constant.Empty };
+  let value = getCookie(Constant.SessionKey);
+  if (value) {
+    let parts = value.split(Constant.Separator);
+    if (parts.length === 5 && output.ts - num(parts[1]) < Setting.SessionTimeout) {
+      output.session = parts[0];
+      output.ts = num(parts[1]);
+      output.count = num(parts[2]) + 1;
+      output.upgrade = num(parts[3]);
+      output.upload = parts[4];
     }
   }
   return output;
@@ -129,11 +142,12 @@ function num(string: string, base: number = 10): number {
 }
 
 function user(): string {
-  let id = cookie(Constant.CookieKey);
-  return id && id.length > 0 ? id : shortid();
+  let id = getCookie(Constant.CookieKey);
+  // Splitting and looking up first part for forward compatibility, in case we wish to store additional information in a cookie
+  return id && id.length > 0 ? id.split(Constant.Separator)[0] : shortid(); 
 }
 
-function cookie(key: string): string {
+function getCookie(key: string): string {
   if (supported(document, Constant.Cookie)) {
     let cookies: string[] = document.cookie.split(Constant.Semicolon);
     if (cookies) {
@@ -146,4 +160,13 @@ function cookie(key: string): string {
     }
   }
   return null;
+}
+
+function setCookie(key: string, value: string, time: number): void {
+  if (config.track && supported(document, Constant.Cookie)) {
+    let expiry = new Date();
+    expiry.setDate(expiry.getDate() + time);
+    let expires = expiry ? Constant.Expires + expiry.toUTCString() : Constant.Empty;
+    document.cookie = `${key}=${value}${Constant.Semicolon}${expires}${Constant.Path}`;
+  }
 }
--- a/packages/clarity-js/src/data/metadata.ts
+++ b/packages/clarity-js/src/data/metadata.ts
@@ -130,8 +130,10 @@ function shortid(): string {
 
 function session(): Session {
   let output: Session = { session: shortid(), ts: Math.round(Date.now()), count: 1, upgrade: BooleanFlag.False, upload: Constant.Empty };
+  // In subsequent versions we will start reading cookies: getCookie(Constant.SessionKey)
+  // For backward compatibility, we will continue reading from session storage in this version
   let legacy = supported(window, Constant.SessionStorage) ? sessionStorage.getItem(Constant.SessionKey) : null; // For backward compatibility
-  let value = getCookie(Constant.SessionKey) || legacy;
+  let value = legacy;
   if (value) {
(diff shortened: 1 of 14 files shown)


packages/clarity-js/types/data.d.ts
```diff
@@ -92,7 +92,9 @@ export const enum Dimension {
     ProductCategory = 11,
     ProductSku = 12,
     ProductCurrency = 13,
-    ProductCondition = 14
+    ProductCondition = 14,
+    TabId = 15,
```
toby-walker, line 96: Shouldn't this go to the envelope? Doesn't seem like it belongs here.

packages/clarity-js/types/data.d.ts
```diff
@@ -134,7 +136,8 @@ export const enum BooleanFlag {
 
 export const enum Setting {
     Expire = 365, // 1 Year
-    SessionTimeout = 30 * Time.Minute, // 10 minutes
+    SessionExpire = 1, // 1 Day
+    SessionTimeout = 30 * Time.Minute, // 30 minutes
```
toby-walker, line 140: Why shrink?

packages/clarity-js/types/core.d.ts
```diff
@@ -102,6 +102,7 @@ export interface Config {
     regions?: Regions;
     metrics?: Metrics;
     cookies?: string[];
+    server?: string;
```
toby-walker, line 105: What if this is empty/undefined?

packages/clarity-js/src/data/upload.ts
```diff
@@ -126,8 +126,8 @@ function stringify(encoded: EncodedPayload): string {
 
 function send(payload: string, sequence: number, last: boolean): void {
     // Upload data if a valid URL is defined in the config
-    if (typeof config.upload === Constant.String) {
-        const url = config.upload as string;
+    if (typeof config.upload === Constant.String && config.server) {
```
toby-walker, line 129: Thsi seems like it wouldn't be backward compatible. Will stop uploading everything on older config.

packages/clarity-js/src/data/metric.ts
```diff
@@ -5,7 +5,9 @@ import encode from "./encode";
 export let data: MetricData = null;
 export let updates: MetricData = null;
 let metricMap: WeakMap<Node, string> = null; // Maps metric nodes => innerText
-const regex = /[^0-9\.]/g;
+const numberRegex = /[^0-9\.,]/g;
+const digitsRegex = /[^0-9\.]/g;
```
toby-walker, line 9: What's the diff between a number regex and a digits regex in semantics? I see the difference in regex, but not clear what you intend from names.

packages/clarity-js/src/data/metric.ts
```diff
@@ -5,7 +5,9 @@ import encode from "./encode";
 export let data: MetricData = null;
 export let updates: MetricData = null;
 let metricMap: WeakMap<Node, string> = null; // Maps metric nodes => innerText
-const regex = /[^0-9\.]/g;
```
toby-walker, line 8: NAive JS question: Is it more efficient to put the consts here rahter than in the function that use them (where it's easier to read the code as they are next to each other)?

packages/clarity-js/src/data/metadata.ts
```diff
@@ -19,19 +19,29 @@ export function start(): void {
...
-    sessionId: s.id,
+    sessionId: s.session,
     pageNum: s.count
   }
 
+  // For backward compatibility; remove in future iterations (v0.6.11)
+  if (config.upload && typeof config.upload === Constant.String && (config.upload as string).indexOf(Constant.HTTPS) === 0) {
+    let url = config.upload as string;
+    config.server = url.substr(0, url.indexOf("/", Constant.HTTPS.length));
+    config.upload = config.server.length > 0 && config.server.length < url.length ? url.substr(config.server.length + 1) : url;
```
toby-walker, line 30: >&& [](start = 45, length = 2)

All this string munging is hard to follow. What are you doing here? Can you add a comment about what you this is for?

toby-walker, line 30: Why do you truncate the URL to the server name only in this?

packages/clarity-js/src/data/metadata.ts
```diff
@@ -71,17 +81,27 @@ export function consent(): void {
 }
 
 export function clear(): void {
-  // Clear any stored information in the session so we can restart fresh the next time
-  if (supported(window, Constant.SessionStorage)) { sessionStorage.removeItem(Constant.StorageKey); }
+  // Clear any stored information in the session cookie so we can restart fresh the next time
```
toby-walker, line 84: >session [](start = 41, length = 7)

session cookie is always a bit amgiuous between (1) a permanent cookie that hold the session id and (2) a session-scoped (lifetime) cookie. Reword to clarify which you mean.

packages/clarity-js/src/data/metadata.ts
```diff
@@ -71,17 +81,27 @@ export function consent(): void {
...
+  setCookie(Constant.SessionKey, Constant.Empty, 0);
+}
+
+function tab(): string {
+  let id = shortid();
+  if (supported(window, Constant.SessionStorage)) {
+    let value = sessionStorage.getItem(Constant.TabKey);
+    id = value && value.indexOf(Constant.Separator) < 0 ? value : id;
```
toby-walker, line 92: >value [](start = 18, length = 5)

What's the separator for if this doesn't use it. Backwards compatible?

toby-walker:
What happens in this scenario:

1) Old versoin from one DC
2) This version from another DC
3) Back to old version from another DC

If the DC rollout hasn't fniished so tha we have the same browser "flapping" between old and new versions?

packages/clarity-js/src/data/metadata.ts
```diff
@@ -90,13 +110,7 @@ function supported(target: Window | Document, api: string): boolean {
...
 
 function track(): void {
-  if (config.track && supported(document, Constant.Cookie)) {
-    let expiry = new Date();
-    expiry.setDate(expiry.getDate() + Setting.Expire);
-    let expires = expiry ? Constant.Expires + expiry.toUTCString() : Constant.Empty;
-    let value = `${data.userId}${Constant.Semicolon}${expires}${Constant.Path}`;
-    document.cookie = Constant.CookieKey + Constant.Equals + value;
-  }
+  setCookie(Constant.CookieKey, data.userId, Setting.Expire);
```
toby-walker, line 113: Don't we lose the structure of the cookie this way?

packages/clarity-js/src/data/metadata.ts
```diff
@@ -108,17 +122,16 @@ function shortid(): string {
...
-    let value = sessionStorage.getItem(Constant.StorageKey);
-    if (value && value.indexOf(Constant.Separator) >= 0) {
-      let parts = value.split(Constant.Separator);
-      if (parts.length === 5 && output.ts - num(parts[1]) < Setting.SessionTimeout) {
-        output.id = parts[0];
-        output.count = num(parts[2]) + 1;
-        output.upgrade = num(parts[3]);
-        output.upload = parts[4];
-      }
+  let output: Session = { session: shortid(), ts: Math.round(Date.now()), count: 1, upgrade: BooleanFlag.False, upload: Constant.Empty };
```
toby-walker, line 125: What happened to the condition on config.track?

packages/clarity-js/src/data/metadata.ts
```diff
@@ -129,11 +142,12 @@ function num(string: string, base: number = 10): number {
 }
 
 function user(): string {
-  let id = cookie(Constant.CookieKey);
-  return id && id.length > 0 ? id : shortid();
+  let id = getCookie(Constant.CookieKey);
+  // Splitting and looking up first part for forward compatibility, in case we wish to store additional information in a cookie
+  return id && id.length > 0 ? id.split(Constant.Separator)[0] : shortid();
```
toby-walker, line 147: >Separator [](start = 49, length = 9)

CookieSeparator to make it clear what is a separator for (shortened like CookieSep or something is fine).

toby-walker (changes requested):
:clock1:

packages/clarity-js/types/core.d.ts
sarveshnagpal: Then that's fine - it will be handled like how we handled empty upload (or upload callback) scenario.

packages/clarity-js/types/data.d.ts
sarveshnagpal: Not expanding... only updating the comment to reflect the true value so comment stays accurate.

packages/clarity-js/src/data/metadata.ts
```diff
@@ -19,19 +19,29 @@ export function start(): void {
...
+    sessionId: s.session,
     pageNum: s.count
   }
 
+  // For backward compatibility; remove in future iterations (v0.6.11)
+  if (config.upload && typeof config.upload === Constant.String && (config.upload as string).indexOf(Constant.HTTPS) === 0) {
+    let url = config.upload as string;
+    config.server = url.substr(0, url.indexOf("/", Constant.HTTPS.length));
+    config.upload = config.server.length > 0 && config.server.length < url.length ? url.substr(config.server.length + 1) : url;
+  }
```
swaathee, line 31: any reason why we cant just keep track of clarity-js version in the session storage for the backward forward compatibility - if the version is not-updated you can send to old session DC else send to new one (and reset session?)

packages/clarity-js/types/data.d.ts
sarveshnagpal: Thought about it and decided to not include it in the envelope because it has no bearing on aggregation. It's a hint for visualization - if we choose to use it later.

packages/clarity-js/src/data/metadata.ts
```diff
@@ -108,17 +122,16 @@ function shortid(): string {
...
-        output.count = num(parts[2]) + 1;
-        output.upgrade = num(parts[3]);
-        output.upload = parts[4];
-      }
+  let output: Session = { session: shortid(), ts: Math.round(Date.now()), count: 1, upgrade: BooleanFlag.False, upload: Constant.Empty };
+  let value = getCookie(Constant.SessionKey);
+  if (value) {
+    let parts = value.split(Constant.Separator);
+    if (parts.length === 5 && output.ts - num(parts[1]) < Setting.SessionTimeout) {
+      output.session = parts[0];
```
swaathee, line 130: >session [](start = 13, length = 7)

how are we dealing with overlapping pageviews across tabs?

packages/clarity-js/src/data/metric.ts
sarveshnagpal: I will rename the variables to be more clear. In one case we remove commas (for en fallback if Intl.NumberFormat is not present) and in another we keep the comma to handle it properly respecting page language.

packages/clarity-js/src/data/metric.ts
sarveshnagpal: It's a good question. It's possible things have improved over time and browsers are smart about handling it internally now but here's a link that provides some additional info: https://stackoverflow.com/questions/8814009/how-often-does-javascript-recompile-regex-literals-in-functions

packages/clarity-js/src/data/metadata.ts
sarveshnagpal: Now that we use cookie to store session state - it seemed wasteful to cache the complete URL. Especially since cookies are sent along with every network requests. This is backward compatibility code. And in future, we can simplify by sending server variable once and then report or upload url can be relative paths.

packages/clarity-js/src/data/metadata.ts
sarveshnagpal: Sounds good. Will reword.

packages/clarity-js/src/data/metadata.ts
```diff
@@ -71,17 +81,27 @@ export function consent(): void {
...
 
 export function save(): void {
   let ts = Math.round(Date.now());
   let upgrade = config.lean ? BooleanFlag.False : BooleanFlag.True;
   let upload = typeof config.upload === Constant.String ? config.upload : Constant.Empty;
   if (upgrade && callback) { callback(data, !config.lean); }
-  if (config.track && supported(window, Constant.SessionStorage)) {
-    sessionStorage.setItem(Constant.StorageKey, [data.sessionId, ts, data.pageNum, upgrade, upload].join(Constant.Separator));
+  if (config.track && supported(document, Constant.Cookie)) {
+    setCookie(Constant.SessionKey, [data.sessionId, ts, data.pageNum, upgrade, upload].join(Constant.Separator), Setting.SessionExpire);
```
swaathee, line 104: this check is present in the setcookie function anyway

packages/clarity-js/src/data/metadata.ts
```diff
@@ -71,17 +81,27 @@ export function consent(): void {
...
 export function clear(): void {
-  // Clear any stored information in the session so we can restart fresh the next time
-  if (supported(window, Constant.SessionStorage)) { sessionStorage.removeItem(Constant.StorageKey); }
+  // Clear any stored information in the session cookie so we can restart fresh the next time
+  setCookie(Constant.SessionKey, Constant.Empty, 0);
+}
+
+function tab(): string {
+  let id = shortid();
+  if (supported(window, Constant.SessionStorage)) {
```
swaathee, line 90: >supported [](start = 6, length = 9)

should we have config.track check here?

packages/clarity-js/src/data/metadata.ts
sarveshnagpal: Sorry not sure if I fully follow. I'm keeping the same format for user cookie as before. Did I miss something here?

packages/clarity-js/src/data/metadata.ts
sarveshnagpal: I will change it to Pipe which is what the character is.

packages/clarity-js/src/data/metadata.ts
sarveshnagpal: It's handled in the visualization right now.

packages/clarity-js/src/data/metadata.ts
sarveshnagpal: Good point, will remove, thanks.

packages/clarity-js/src/data/metadata.ts
sarveshnagpal: I thought tracking tab is not the same as tracking user or session. But, I do see your point. Let me add the check.

packages/clarity-js/src/data/metric.ts
sarveshnagpal: Done.

packages/clarity-js/src/data/metadata.ts
sarveshnagpal: I'll add comments to explain it more. I am hoping we can remove this bit with the next version.

sarveshnagpal: Done

packages/clarity-js/src/data/metadata.ts
sarveshnagpal: Done

packages/clarity-js/src/data/metadata.ts
sarveshnagpal: Done

packages/clarity-js/src/data/metadata.ts
sarveshnagpal: Done

packages/clarity-js/src/data/metadata.ts
sarveshnagpal: Done

swaathee (approved):
:shipit:

sarveshnagpal:
Synced offline with Swaathee and verified the changes work fine in all permutations (that we could think of). 

---
In reply to: [816971978](https://github.com/microsoft/clarity/pull/127#issuecomment-816971978) [](ancestors = 816971978)

sarveshnagpal (commented):
Tested the changes with multiple permutations.

PR #128, opened 2021-04-11
JSON+LD fixes and moving sessionization to use cookies

sarveshnagpal:
- Better handle unexpected payloads in JSON+LD
- Use cookie instead of session storage for session - so we can track cross tabs
- Fall back to sendBeacon when XHR return status code = 0 (on page getting unloaded). Impacts Safari.

### DIFF
--- a/packages/clarity-js/src/data/upload.ts
+++ b/packages/clarity-js/src/data/upload.ts
@@ -163,14 +163,18 @@ function check(xhr: XMLHttpRequest, sequence: number, last: boolean): void {
         if ((xhr.status < 200 || xhr.status > 208) && transitData.attempts <= Setting.RetryLimit) {
             // We re-attempt in all cases except two: 
             //     0: Indicates the browser has not put the request on the wire and therefore we need to attempt sendBeacon API before giving up
-            //   400: Indicates the server has rejected the response for bad payload and therefore we terminate the session
-            switch (xhr.status) {
+            //   4XX: Indicates the server has rejected the response for bad payload and therefore we terminate the session
+            if (xhr.status === 0) {
                 // The observed behavior is that Safari will terminate pending XHR requests with status code 0
                 // if the user navigates away from the page. In these cases, we fallback to the else clause and lose the data
                 // By explicitly handing status code 0 we attempt to try a different transport (sendBeacon vs. XHR) before giving up.
-                case 0: send(transitData.data, sequence, true); break;
-                case 400: limit.trigger(Check.Server); break;
-                default: send(transitData.data, sequence, last); break;
+                send(transitData.data, sequence, true);
+            } else if (xhr.status >= 400 && xhr.status < 500) {
+                // Anytime we receive a 4XX response from the server, we bail out instead of trying again
+                limit.trigger(Check.Server);
+            } else {
+                // In all other cases, re-attempt sending the same data
+                send(transitData.data, sequence, last);
             }
         } else {
             track = { sequence, attempts: transitData.attempts, status: xhr.status };
--- a/packages/clarity-js/src/data/upload.ts
+++ b/packages/clarity-js/src/data/upload.ts
@@ -124,7 +124,7 @@ function stringify(encoded: EncodedPayload): string {
     return encoded.p.length > 0 ? `{"e":${encoded.e},"a":${encoded.a},"p":${encoded.p}}` : `{"e":${encoded.e},"a":${encoded.a}}`;
 }
 
-function send(payload: string, sequence: number, last: boolean): void {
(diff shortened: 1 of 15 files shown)


packages/clarity-js/src/data/upload.ts
```diff
@@ -124,7 +124,7 @@ function stringify(encoded: EncodedPayload): string {
     return encoded.p.length > 0 ? `{"e":${encoded.e},"a":${encoded.a},"p":${encoded.p}}` : `{"e":${encoded.e},"a":${encoded.a}}`;
 }
 
-function send(payload: string, sequence: number, last: boolean): void {
+function send(payload: string, sequence: number, beacon: boolean): void {
```
toby-walker, line 127: >beacon [](start = 49, length = 6)

Suggest you do this for all 4xx rather than just 400.

packages/clarity-js/src/layout/schema.ts
```diff
@@ -39,7 +41,7 @@ export function ld(json: any): void {
                     dimension.log(Dimension.ProductCondition, json[JsonLD.ItemCondition]);
                     dimension.log(Dimension.ProductCurrency, json[JsonLD.PriceCurrency]);
                     dimension.log(Dimension.ProductSku, json[JsonLD.Sku]);
-                    metric.max(Metric.ProductPrice, num(json[JsonLD.Price]));
+                    metric.max(Metric.ProductPrice, num(json[JsonLD.Price], Setting.PriceScale));
```
toby-walker, line 44: >PriceScale [](start = 84, length = 10)

Will require a server-side change?

packages/clarity-js/src/layout/schema.ts
```diff
@@ -51,6 +53,12 @@ export function ld(json: any): void {
     }
 }
 
-function num(input: string): number {
-    return input ? Math.round(parseFloat(input)) : null;
+function num(input: string | number, scale: number = 1): number {
+    if (input !== null) {
+        switch (typeof input) {
+            case Constant.Number: return Math.round((input as number) * scale);
+            case Constant.String: return Math.round(parseFloat((input as string).replace(digitsRegex, Constant.Empty)) * scale);
```
toby-walker, line 60: Shouldn't this be done in another higher layer if we expect this it have non-numeric parts?

toby-walker (approved):
:shipit:

packages/clarity-js/src/layout/schema.ts
```diff
@@ -24,7 +26,7 @@ export function ld(json: any): void {
                     break;
                 case JsonLD.AggregateRating:
                     if (json[JsonLD.RatingValue]) {
-                        metric.max(Metric.RatingValue, parseFloat(json[JsonLD.RatingValue]) * Setting.RatingScale);
+                        metric.max(Metric.RatingValue, num(json[JsonLD.RatingValue], Setting.RatingScale));
```
toby-walker, line 29: IMPORTANT: Please do not do this in this PR for reasons I put in our convsation. Instead can you either leave it unscaled as now or create a seprate metric that has the scaled value and send both (for now) until we migrate over?

toby-walker (changes requested):
:clock1:

packages/clarity-js/src/data/upload.ts
sarveshnagpal: Done

packages/clarity-js/src/data/upload.ts
```diff
@@ -161,19 +161,26 @@ function check(xhr: XMLHttpRequest, sequence: number, last: boolean): void {
...
-            } else { send(transitData.data, sequence, last); }
+            // We re-attempt in all cases except two: 
+            //     0: Indicates the browser has not put the request on the wire and therefore we need to attempt sendBeacon API before giving up
+            //   400: Indicates the server has rejected the response for bad payload and therefore we terminate the session
+            switch (xhr.status) {
+                // The observed behavior is that Safari will terminate pending XHR requests with status code 0
+                // if the user navigates away from the page. In these cases, we fallback to the else clause and lose the data
+                // By explicitly handing status code 0 we attempt to try a different transport (sendBeacon vs. XHR) before giving up.
+                case 0: send(transitData.data, sequence, true); break;
+                case 400: limit.trigger(Check.Server); break;
```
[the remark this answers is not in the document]
sarveshnagpal: Synced offline - will understand this better & tackle this in a separate change.

packages/clarity-js/src/layout/schema.ts
sarveshnagpal: Rolled it back per our offline conversation.

toby-walker (approved):
:shipit:

PR #132, opened 2021-05-05
Do not track playback until we send back more than 0 dom bytes

swaathee:
Playback is set from upload when we send > 0 dom bytes - removing the metric when calling upgrade, since we could upgrade but the client might not send any playback data


toby-walker (approved):
:shipit:

sarveshnagpal (approved):

PR #137, opened 2021-05-19
Performance improvements and simplifying regions

sarveshnagpal:
- Refactor region logic
- Consolidate code to extra data from DOM
- Updating data schema (breaking changes)
- Adding backward compatibility to decoding layer
- Updating visualization to read new format


packages/clarity-decode/src/layout.ts
```diff
@@ -20,12 +20,12 @@ export function decode(tokens: Data.Token[]): LayoutEvent {
...
             for (let i = 2; i < tokens.length; i += 3) {
-                // For backward compatibility since 0.6.4, we extract visibility signal from the "box" field if it is present
-                let legacy = Array.isArray(tokens[i+1]) && (tokens[i+1] as number[]).length === 5;
+                // For backward compatibility since 0.6.12, map visibility to state
+                let legacy = typeof tokens[i + 1] === Constant.Number;
                 let region: Layout.RegionData = {
                     id: tokens[i] as number,
-                    visible: legacy ? tokens[i + 1][4] as number : tokens[i + 1] as number,
-                    region: tokens[i + 2] as string
+                    name: legacy ? tokens[i + 2] as string : tokens[i + 1] as string,
```
swaathee, line 27: >tokens [](start = 35, length = 6)

why not keep it consistent - i + 1 is state, i + 2 is name ?

packages/clarity-js/src/layout/extract.ts
```diff
@@ -0,0 +1,64 @@
...
+const digitsRegex = /[^0-9\.]/g;
+const digitsWithCommaRegex = /[^0-9\.,]/g;
+
+export function regions(root: ParentNode, value: Region[]): void {
+    for (let v of value) {
+        let valid = true;
+        let filter = v.length === 4 ? v[2] : null;
+        switch (filter) {
+            case RegionFilter.Url: valid = !!top.location.href.match(new RegExp(v[3])); break;
+            case RegionFilter.Javascript: valid = !!evaluate(v[3]); break;
```
swaathee, line 16: >Javascript [](start = 30, length = 10)

how would we handle cases where the js variable is set after an element is parsed?

packages/clarity-js/src/layout/mutation.ts
```diff
@@ -145,6 +150,42 @@ async function process(): Promise<void> {
...
+    history[key][0] = inactive ? history[key][0] + 1 : 1;
+    // Return updated mutation type based on if we have already hit the threshold or not
+    if (history[key][0] === Setting.MutationSuspendThreshold) {
+      // Store a reference to removedNodes so we can process them later
+      history[key][1] = m.removedNodes;
+      return Constant.Suspend;
+    } else if (history[key][0] > Setting.MutationSuspendThreshold) { return Constant.Empty; }
+  }
+  return m.type;
+}
```
swaathee, line 175: is there a chance type is not null when target is null ?

packages/clarity-js/src/layout/region.ts
```diff
@@ -44,13 +44,19 @@ export function exists(node: Node): boolean {
...
-export function track(id: number): void {
-    // Do not send another entry if we are already sending it
+export function track(id: number, event: Event): void {
     let node = dom.getNode(id);
-    if (updates.indexOf(id) <= 0 && regionMap.has(node)) {
-        regions[id] = id in regions ? regions[id] : { id, visible: BooleanFlag.False, region: regionMap.get(node) };
-        data.push(regions[id]);
-    }
+    let d = id in regions ? regions[id] : { id, name: regionMap.get(node), state: RegionState.Rendered };
+    let state = event === Event.Click && (d.state !== RegionState.Clicked && d.state !== RegionState.Input) ? RegionState.Clicked : d.state;
```
swaathee, line 50: >d.state !== RegionState.Clicked [](start = 42, length = 31)

do we need this check? we would anyway reassign the same value

packages/clarity-js/src/layout/extract.ts
```diff
@@ -0,0 +1,64 @@
...
+        let scale = v.length === 4 ? v[3] : 1;
+        switch (source) {
+            case Extract.Text:  root.querySelectorAll(v[2]).forEach(e => { metric.max(v[0], num((e as HTMLElement).innerText, scale)); }); break;
+            case Extract.Attribute: root.querySelectorAll(`[${v[2]}]`).forEach(e => { metric.max(v[0], num(e.getAttribute(v[2]), scale, false)); }); break;
+            case Extract.Javascript: metric.max(v[0], evaluate(v[2], Constant.Number) as number); break;
+        }
+    }
+}
+
+function evaluate(variable: string, type: string = null, base: Object = window): any {
```
swaathee, line 34: >evaluate [](start = 9, length = 8)

can you add comments to explain the purpose of this

packages/clarity-js/src/layout/mutation.ts
```diff
@@ -145,6 +150,42 @@ async function process(): Promise<void> {
...
+    let key = [value.selector, m.attributeName, m.addedNodes ? m.addedNodes.length : 0, m.removedNodes ? m.removedNodes.length : 0].join();
+    // Initialize an entry if it doesn't already exist
+    history[key] = key in history ? history[key] : [0];
+    // Lookup any pending nodes queued up for removal, and process them now if we had suspended a mutation before
+    if (inactive === false && history[key][0] >= Setting.MutationSuspendThreshold) { processNodeList(history[key][1], Source.ChildListRemove, timer); }
+    // Update the counter
+    history[key][0] = inactive ? history[key][0] + 1 : 1;
+    // Return updated mutation type based on if we have already hit the threshold or not
+    if (history[key][0] === Setting.MutationSuspendThreshold) {
+      // Store a reference to removedNodes so we can process them later
```
swaathee, line 169: should we check if the node is inactive here? it looks like we are not processing the current changes to the element during the active period.

swaathee (changes requested):
:clock1:

packages/clarity-js/src/layout/extract.ts
sarveshnagpal: It is triggered on mutations too so we will have more opportunities to look it up.

packages/clarity-decode/src/layout.ts
sarveshnagpal: Good point. will update.

---
In reply to: [643631103](https://github.com/microsoft/clarity/pull/137#discussion_r643631103) [](ancestors = 643631103)

packages/clarity-js/src/layout/extract.ts
sarveshnagpal: We process this after every mutation (see: dom.ts) - so we will be fine if variable is introduced later in the page.

---
In reply to: [643631150](https://github.com/microsoft/clarity/pull/137#discussion_r643631150) [](ancestors = 643631150)

packages/clarity-js/src/layout/extract.ts
sarveshnagpal: done, thanks.

---
In reply to: [643631761](https://github.com/microsoft/clarity/pull/137#discussion_r643631761) [](ancestors = 643631761)

packages/clarity-js/src/layout/mutation.ts
sarveshnagpal: We already do so within processNode function:
// Do not track this change if we are attempting to remove a node before discovering it
    if (source === Source.ChildListRemove && dom.has(node) === false) { return child; }

If the node is destroyed before we are able to process "remove" on it - we do handle it gracefully. 

---
In reply to: [643633409](https://github.com/microsoft/clarity/pull/137#discussion_r643633409) [](ancestors = 643633409)

packages/clarity-js/src/layout/mutation.ts
sarveshnagpal: It may be possible but the code path doesn't change from what we had earlier. If target was null then we will continue processing the mutation like we did (without maintaining history) and no new change in behavior will be introduced.

---
In reply to: [643631705](https://github.com/microsoft/clarity/pull/137#discussion_r643631705) [](ancestors = 643631705)

packages/clarity-js/src/layout/region.ts
sarveshnagpal: Good point - removed that particular check. Now: Event.Click && d.state !== Interaction.Input ? Interaction.Clicked : d.state;

---
In reply to: [643631746](https://github.com/microsoft/clarity/pull/137#discussion_r643631746) [](ancestors = 643631746)

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "clarity",
   "private": true,
-  "version": "0.6.14",
```
toby-walker, line 4: AT some point we should start using the other parts of hte sesmantic version if the data contract breaks. This seems like it might be a 0.7.x release.

packages/clarity-decode/src/interaction.ts
```diff
@@ -32,8 +32,7 @@ export function decode(tokens: Data.Token[]): InteractionEvent {
                 context: tokens[9] as number,
                 text: tokens[10] as string,
                 link: tokens[11] as string,
-                hash: tokens[12] as string,
-                region: tokens.length > 13 ? tokens[13] as number : null
```
toby-walker, line 36: I don't think we should tkae this out as we'll lose the association of time and timeline for activiites. Let' discuss this offline.

packages/clarity-devtools/src/config.ts
```diff
@@ -0,0 +1,16 @@
+import { Core, Data, Layout } from "clarity-js";
+
+export default function(): Core.Config {
+    return {
+        regions: [
+            [Data.Region.Header, "header", 1, Layout.Constant.DevHook], /* 1: Javascript Filter */
+            [Data.Region.Footer, "footer", 0, "product"], /* 0: Url */
+            [Data.Region.Navigation, "nav"]
+        ],
```
toby-walker, line 9: What's this file for?

toby-walker:
If it's easy can you collect rtl as a dimension?

---
In reply to: [673760926](https://github.com/microsoft/clarity/pull/137#issuecomment-673760926) [](http://example.com/codeflow?ancestors=673760926,673760926)

packages/clarity-js/src/layout/extract.ts
```diff
@@ -0,0 +1,67 @@
...
+const formatRegex = /1/g;
+const digitsRegex = /[^0-9\.]/g;
+const digitsWithCommaRegex = /[^0-9\.,]/g;
+
+export function regions(root: ParentNode, value: Region[]): void {
+    for (let v of value) {
+        let valid = true;
+        let filter = v.length === 4 ? v[2] : null;
+        switch (filter) {
+            case RegionFilter.Url: valid = !!top.location.href.match(new RegExp(v[3])); break;
```
toby-walker, line 15: SHould we cache these regexs instead of creating them every time here?

packages/clarity-js/src/layout/extract.ts
```diff
@@ -0,0 +1,67 @@
...
+import * as region from "@src/layout/region";
+
+const formatRegex = /1/g;
+const digitsRegex = /[^0-9\.]/g;
+const digitsWithCommaRegex = /[^0-9\.,]/g;
+
+export function regions(root: ParentNode, value: Region[]): void {
+    for (let v of value) {
+        let valid = true;
+        let filter = v.length === 4 ? v[2] : null;
```
toby-walker, line 13: Suggestion: CAn you use a desstrcutring assignment to get the components of the Region. It would be something like

const  [a, b, c, d] = v 

where a,b, c, d are the right names for hte lements of the config.

packages/clarity-js/src/layout/extract.ts
```diff
@@ -0,0 +1,67 @@
...
+    try {
+        // Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat
+        let lang = document.documentElement.lang;
+        if (Intl && Intl.NumberFormat && lang && localize) {
+            text = text.replace(digitsWithCommaRegex, Constant.Empty);
+            // Infer current group and decimal separator from current locale
+            let group = Intl.NumberFormat(lang).format(11111).replace(formatRegex, Constant.Empty);
+            let decimal = Intl.NumberFormat(lang).format(1.1).replace(formatRegex, Constant.Empty);
+            
+            // Prase number using inferred group and decimal separators
```
toby-walker, line 58: Typo for parse

packages/clarity-js/src/layout/extract.ts
```diff
@@ -0,0 +1,67 @@
...
+            return Math.round(parseFloat(text
+                .replace(new RegExp('\\' + group, 'g'), Constant.Empty)
+                .replace(new RegExp('\\' + decimal), Constant.Dot)
+            ) * scale);
+        }
+        // Fallback to en locale
+        return Math.round(parseFloat(text.replace(digitsRegex, Constant.Empty)) * scale);
+    } catch { return 0; }
```
toby-walker, line 66: CAn we not put a number if we fail here. Cleaner to check where this is called and return undefined if we can't parse it.

packages/clarity-js/types/core.d.ts
```diff
@@ -3,6 +3,8 @@ import { Metadata, Payload, Token } from "./data";
 type TaskFunction = () => Promise<void>;
 type TaskResolve = () => void;
 type UploadCallback = (data: string) => void;
+type Region = [number, string, number?, string?];
```
toby-walker, line 6: Use the correct typing here. I think the second (?) is a RegionFilter and for MEtric the second is a Extract.

packages/clarity-js/types/data.d.ts
```diff
@@ -97,6 +97,12 @@ export const enum Dimension {
     PageLanguage = 16
 }
 
+export const enum Region {
```
toby-walker, line 100: Why are these needed and what are they for?

packages/clarity-js/types/layout.d.ts
```diff
@@ -87,6 +98,8 @@ export const enum JsonLD {
 
 export const enum Setting {
     LookAhead = 33, // 33ms
+    MutationSuspendThreshold = 10, // Stop listening for mutations after hitting a threshold
+    MutationActivePeriod = 3000 // Let mutations continue as normal during active periods of user interactions
```
toby-walker, line 102: >3000 [](http://example.com/codeflow?start=27&length=4)

What are the time units for these? ms or seconds?

packages/clarity-js/types/layout.d.ts
```diff
@@ -151,8 +168,8 @@ export interface DocumentData {
 
 export interface RegionData {
     id: number;
-    visible: number;
-    region: string;
+    state: Interaction;
```
toby-walker, line 171: >Interaction [](http://example.com/codeflow?start=11&length=11)

Can we call this InteractionState as it's weird to call a "state" property an Interaction type.

package.json
sarveshnagpal: It is backward compatible. We have been bumping minor version (0.7) when we make breaking changes.

packages/clarity-decode/src/interaction.ts
sarveshnagpal: We reviewed this internally few weeks back. We can discuss this offline but this is part of the simplication and reducing chatter over wire

packages/clarity-devtools/src/config.ts
sarveshnagpal: For dev tools that we package with clarity to test out region functionality.

packages/clarity-js/types/data.d.ts
sarveshnagpal: This is proof of concept. We will now deal with pre-defined regions (as part of offline spec). So once we have the list of regions Clarity supports, it will show up as const enum here (similar to metrics and dimensions).

swaathee, line 100: >Region [](http://example.com/codeflow?start=18&length=6)

I thought we will not send down the region list in code - and this would be a mapping maintained by the server

packages/clarity-visualize/src/heatmap.ts
```diff
@@ -72,8 +142,8 @@ export function click(activity: Activity): void {
 
 function overlay(): HTMLCanvasElement {
     // Create canvas for visualizing heatmap
-    let doc = state.player.contentDocument;
-    let win = state.player.contentWindow;
+    let doc = state.window.document;
+    let win = state.window;
     let de = doc.documentElement;
     let canvas = doc.getElementById(Constant.HeatmapCanvas) as HTMLCanvasElement;
     if (canvas === null) {
```
swaathee, line 149: should we check for options.canvas here?

swaathee (approved):
:shipit:

packages/clarity-js/types/layout.d.ts
```diff
@@ -10,6 +11,13 @@ export const enum Source {
     CharacterData
 }
 
+export const enum Interaction {
+    Rendered = 0,
+    Visible = 1,
+    Clicked = 2,
+    Input = 3
```
swaathee, line 18: can you define these with gaps in between the numbers?

packages/clarity-js/src/layout/mutation.ts
```diff
@@ -145,6 +150,42 @@ async function process(): Promise<void> {
...
 }
 
+function track(m: MutationRecord, timer: Metric): string {
+  let value = m.target ? dom.get(m.target.parentNode) : null;
+  if (value) {
+    let inactive = time() > activePeriod;
+    // We use selector, instead of id, to determine the key because in some cases
+    // repeated mutations can cause elements to be destroyed and then recreated as new DOM nodes
+    // In those cases, IDs will change however the selector (which is relative to DOM xPath) remains the same
+    let key = [value.selector, m.attributeName, m.addedNodes ? m.addedNodes.length : 0, m.removedNodes ? m.removedNodes.length : 0].join();
```
toby-walker, line 160: ARe there any cases they might cause issues? Just applying "how can this break" thinking.

toby-walker: For example a false positive or false negative based on same key when it should be a different key (or vice versa)?

packages/clarity-js/src/layout/mutation.ts
```diff
@@ -145,6 +150,42 @@ async function process(): Promise<void> {
...
+    // Initialize an entry if it doesn't already exist
+    history[key] = key in history ? history[key] : [0];
```
toby-walker, line 162: >0 [](http://example.com/codeflow?start=52&length=1)

What's the [0] represent?

packages/clarity-js/src/layout/mutation.ts
```diff
@@ -145,6 +150,42 @@ async function process(): Promise<void> {
...
+    // Lookup any pending nodes queued up for removal, and process them now if we suspended a mutation before
+    if (inactive === false && history[key][0] >= Setting.MutationSuspendThreshold) { processNodeList(history[key][1], Source.ChildListRemove, timer); }
```
toby-walker, line 164: A JS question: Why prefer this use of an array over an object with two properites (one for the time and the other for hte queued nodes) where the two parts are separated. Is it a performance or memory consideration?

toby-walker, line 164: Nite: Put the {...} on separate lines.

packages/clarity-js/src/layout/mutation.ts
```diff
@@ -145,6 +150,42 @@ async function process(): Promise<void> {
...
+    // Update the counter
+    history[key][0] = inactive ? history[key][0] + 1 : 1;
```
toby-walker, line 166: Mnior suggestion: Cache the history entyr in a local so it make the code simpler to read and avoids all of the histoyr[key] sprinked here. 

let entry = history[key]

Or similar.

packages/clarity-js/src/layout/mutation.ts
```diff
@@ -145,6 +150,42 @@ async function process(): Promise<void> {
...
+    // Return updated mutation type based on if we have already hit the threshold or not
+    if (history[key][0] === Setting.MutationSuspendThreshold) {
+      // Store a reference to removedNodes so we can process them later
```
toby-walker, line 169: CAn you expand the comment to indicate why we only do the removed node and note othert update operatons?

packages/clarity-js/src/layout/region.ts
```diff
@@ -44,13 +44,19 @@ export function exists(node: Node): boolean {
...
 
-export function track(id: number): void {
-    // Do not send another entry if we are already sending it
+export function track(id: number, event: Event): void {
     let node = dom.getNode(id);
-    if (updates.indexOf(id) <= 0 && regionMap.has(node)) {
-        regions[id] = id in regions ? regions[id] : { id, visible: BooleanFlag.False, region: regionMap.get(node) };
-        data.push(regions[id]);
-    }
+    let d = id in regions ? regions[id] : { id, state: Interaction.Rendered, name: regionMap.get(node) };
```
toby-walker, line 49: Nit. What is d an abbrevation for. Can you use a slighlty more meaningful name?

packages/clarity-js/src/layout/region.ts
```diff
@@ -44,13 +44,19 @@ export function exists(node: Node): boolean {
...
+    let state = event === Event.Click && d.state !== Interaction.Input ? Interaction.Clicked : d.state;
```
toby-walker, line 50: >Clicked [](http://example.com/codeflow?start=85&length=7)

To discuss: What if we use a bitmap for hte state of the elmeent so we can track everything?

packages/clarity-js/src/layout/region.ts
```diff
@@ -44,13 +44,19 @@ export function exists(node: Node): boolean {
...
+    state = event === Event.Input && d.state !== Interaction.Input ? Interaction.Input : d.state;
```
toby-walker, line 51: Suggestion: Make these ordered and then just do maxstate = (state,event)

packages/clarity-js/src/layout/region.ts
```diff
@@ -91,14 +97,16 @@ function handler(entries: IntersectionObserverEntry[]): void {
             // However, for larger regions, area of regions could be bigger than viewport and therefore comparison is relative to visible area
             let visible = viewportRatio > Setting.ViewportIntersectionRatio || entry.intersectionRatio > Setting.IntersectionRatio ? BooleanFlag.True : BooleanFlag.False;
             // If the visibility hasn't changed since the last tracked value, ignore this notification
-            if (!(id in regions && regions[id].visible === visible)) {
-                let d = { id, visible, region: regionMap.get(target) };
+            if (!(id in regions && visible && regions[id].state === Interaction.Visible)) {
```
toby-walker, line 100: Minor suggestion is to move the state updating to a single function rather than having the logic spread into mulitple places like this. Something like

if(visibile) {
    updateState(id, Interaction.Visible)
}

Keeps the logic in one place and easier to track it.

toby-walker (changes requested):
:clock1:

package.json
toby-walker: Can you clarify the "it is backward compataible" part? Can discuss in person.

packages/clarity-js/types/data.d.ts
toby-walker: Could yo uadd a comment about the above?

package.json
sarveshnagpal: Discussed offline.

packages/clarity-js/src/layout/extract.ts
sarveshnagpal: Thanks, fixed.

packages/clarity-js/src/layout/region.ts
```diff
@@ -44,13 +44,16 @@ export function exists(node: Node): boolean {
...
 
-export function track(id: number): void {
-    // Do not send another entry if we are already sending it
+export function track(id: number, event: Event): void {
     let node = dom.getNode(id);
-    if (updates.indexOf(id) <= 0 && regionMap.has(node)) {
-        regions[id] = id in regions ? regions[id] : { id, visible: BooleanFlag.False, region: regionMap.get(node) };
-        data.push(regions[id]);
-    }
+    let d = id in regions ? regions[id] : { id, state: InteractionState.Rendered, name: regionMap.get(node) };
```
swaathee, line 49: >InteractionState.Rendered [](http://example.com/codeflow?start=55&length=25)

doesn't rendered get overridden by the interaction state  - will we ever get rendered state?

packages/clarity-js/types/layout.d.ts
sarveshnagpal: done, thanks

packages/clarity-js/types/data.d.ts
sarveshnagpal: Sounds good. Moved it to dev tools to serve as an example and removed it from here for now.

packages/clarity-js/src/layout/extract.ts
sarveshnagpal: Great point, will do.

packages/clarity-js/src/layout/extract.ts
sarveshnagpal: Good point, done.

packages/clarity-js/src/layout/extract.ts
sarveshnagpal: Yes right makes complete sense. No idea why I had it at 0 but fixed it to return null.

packages/clarity-js/types/core.d.ts
sarveshnagpal: Added comments and used references to specific types where possible.

packages/clarity-js/types/data.d.ts
sarveshnagpal: Removed it per Swaathee's comment.

packages/clarity-js/types/layout.d.ts
sarveshnagpal: Added it in comments. The answer is milliseconds.

packages/clarity-js/types/layout.d.ts
sarveshnagpal: Sounds good - done.

packages/clarity-js/src/layout/mutation.ts
sarveshnagpal: We discussed this offline. It errs on the side of letting the mutation pass through if it's not the exactly same signature.

packages/clarity-js/src/layout/mutation.ts
sarveshnagpal: Done

packages/clarity-js/src/layout/mutation.ts
sarveshnagpal: We discussed it offline.

packages/clarity-js/src/layout/mutation.ts
sarveshnagpal: Done

packages/clarity-js/src/layout/region.ts
sarveshnagpal: Changed it to data.

packages/clarity-js/src/layout/region.ts
sarveshnagpal: Done

packages/clarity-js/src/layout/region.ts
sarveshnagpal: Discussed offline and intentionally only want to track the max state.

packages/clarity-js/src/layout/region.ts
sarveshnagpal: also simplified the whole logic per our offline conversation.

packages/clarity-js/src/layout/region.ts
sarveshnagpal: simplified this whole logic.

packages/clarity-js/src/layout/region.ts
sarveshnagpal: Good point & done

packages/clarity-visualize/src/heatmap.ts
sarveshnagpal: yes good point, will do

sarveshnagpal: thought about it more and left it as it is... it's hard to accomplish anything without canvas support in heatmap. will think this over and include it as separate change later.

PR #145, opened 2021-06-14
log document direction

swaathee:


packages/clarity-js/types/data.d.ts
```diff
@@ -94,7 +94,8 @@ export const enum Dimension {
     ProductCurrency = 13,
     ProductCondition = 14,
     TabId = 15,
-    PageLanguage = 16
+    PageLanguage = 16,
+    Direction = 17
```
toby-walker, line 98: Suggest {{Document,/Layout/Text} Direction as Direction is a bit ambiguous (could be pointer direction, et.)

toby-walker (approved):
:shipit:

PR #148, opened 2021-06-23
Using CompressionStream where supported

sarveshnagpal:
- Use CompressionStream API in browsers that support it
- Fallback to using string payload if the request fails for whatever reason

### DIFF
--- a/packages/clarity-js/src/data/compress.ts
+++ b/packages/clarity-js/src/data/compress.ts
@@ -1,0 +1,29 @@
+import { Constant } from "@clarity-types/data";
+
+const supported = Constant.CompressionStream in window;
+
+export default async function(input: string): Promise<Uint8Array> {
+    if (supported) {
+        // Create a readable stream from given input string and
+        // pipe it through text encoder and compression stream to gzip it
+        const stream = new ReadableStream({async start(controller) {
+            controller.enqueue(input);
+            controller.close();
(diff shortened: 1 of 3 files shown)


packages/clarity-js/src/data/upload.ts
```diff
@@ -143,12 +150,21 @@ function send(payload: string, sequence: number, beacon: boolean): void {
...
             xhr.open("POST", url);
             if (sequence !== null) { xhr.onreadystatechange = (): void => { measure(check)(xhr, sequence, beacon); }; }
             xhr.withCredentials = true;
-            xhr.send(payload);
+            if (zipped) {
+                // If we do have valid compressed array, send it with appropriate HTTP headers
+                xhr.setRequestHeader(Constant.ContentEncoding, Constant.Gzip);
+                xhr.send(zipped);
+            } else {
+                // In all other cases, continue sending string back to the server
```
toby-walker, line 165: >all [](http://example.com/codeflow?start=22&length=3)

What about falling back to non-compressed if this fails and you retry it?

packages/clarity-js/src/data/compress.ts
```diff
@@ -0,0 +1,29 @@
+import { Constant } from "@clarity-types/data";
+
+const supported = Constant.CompressionStream in window;
+
+export default async function(input: string): Promise<Uint8Array> {
+    if (supported) {
```
toby-walker, line 6: Suggest you add a try/catch here in case something goes wrong.

toby-walker (approved):
:shipit:

PR #151, opened 2021-07-08
force send dom data on first upgrade

swaathee:


sarveshnagpal:
Integrated changes into #153

PR #156, opened 2021-07-22
Add support for cross sub-domain sessions

sarveshnagpal:
- Identify root domain on the client and set cookies against it instead of current domain
- Backward compatibility with existing cookies to make this change as smooth as possible
- Add cookie version to ensure we can make future changes more smoothly
- Bumping version to 0.6.18


packages/clarity-js/src/data/metadata.ts
```diff
@@ -147,9 +150,23 @@ function num(string: string, base: number = 10): number {
...
 function user(): string {
-  let id = getCookie(Constant.CookieKey);
-  // Splitting and looking up first part for forward compatibility, in case we wish to store additional information in a cookie
-  return id && id.length > 0 ? id.split(Constant.Pipe)[0] : shortid(); 
+  let cookie = getCookie(Constant.CookieKey);
+  if(cookie && cookie.length > 0) {
+    // Splitting and looking up first part for forward compatibility, in case we wish to store additional information in a cookie
+    let parts = cookie.split(Constant.Pipe);
+    let userId = cookie.split(Constant.Pipe)[0];
+    if (parts.length === 2) { return userId; } else {
```
toby-walker, line 158: Minor suggestion: make this >= 2 for forward compat in case we ever add soething else.

packages/clarity-js/src/data/metadata.ts
```diff
@@ -147,9 +150,23 @@ function num(string: string, base: number = 10): number {
...
+      // For backward compatibility introduced in v0.6.18; this else clause can be removed with future iterations
+      let deleted = `${Constant.Semicolon}${Constant.Expires}${(new Date(0)).toUTCString()}${Constant.Path}`;
+      // First, delete current user cookie which might be set on current sub-domain vs. root domain
+      document.cookie = `${Constant.CookieKey}=${deleted}`;
+      // Second, same thing for current session cookie so it can be re-written later with the root domain
+      document.cookie = `${Constant.SessionKey}=${deleted}`;
+      // Return the existing userId, so it can eventually be written with version info later
```
toby-walker, line 165: >userId [](http://example.com/codeflow?start=29&length=6)

What if cookies are not enabled yet? Shold we not do this in that case?

packages/clarity-js/src/data/metadata.ts
```diff
@@ -172,6 +189,30 @@ function setCookie(key: string, value: string, time: number): void {
...
+          if (i < hostname.length - 1) { 
+            // Write the cookie on the current computed top level domain
+            document.cookie = `${cookie}${Constant.Semicolon}${Constant.Domain}${rootDomain}`;
+            // Once written, check if the cookie was set successfully
+            // If yes, no more action required and return from the function. rootDomain is already set
+            // If no, then continue with the for loop
+            if (document.cookie.indexOf(key) >= 0) { return; }
+          }
+        }
+        // Finally, if we were not successful and gone through all the options, play safe and reset rootDomain to be empty
```
toby-walker, line 211: typo play *it* safe

packages/clarity-js/src/data/metadata.ts
```diff
@@ -172,6 +189,30 @@ function setCookie(key: string, value: string, time: number): void {
...
-    document.cookie = `${key}=${value}${Constant.Semicolon}${expires}${Constant.Path}`;
+    let cookie = `${key}=${value}${Constant.Semicolon}${expires}${Constant.Path}`;
+    try {
+      // Attempt to get the root domain only once and fall back to writing cookie on the current domain.
+      if (rootDomain === null) {
+        let hostname = location.hostname ? location.hostname.split(Constant.Dot) : [];
+        // Walk backwards on a domain and attempt to set a cookie, until successful
+        for (let i = hostname.length - 1; i >= 0; i--) {
+          rootDomain = `.${hostname[i]}${rootDomain ? rootDomain : Constant.Empty}`;
+          // We do not wish to attempt writing cookie on the absolute last part, e.g. .com or .net.
```
toby-walker, line 200: >writing [](http://example.com/codeflow?start=39&length=7)

typo: writing

packages/clarity-js/src/data/metadata.ts
```diff
@@ -172,6 +189,30 @@ function setCookie(key: string, value: string, time: number): void {
...
+          // So we start attempting after second-last part, e.g. .domain.com (PASS) or .co.uk (FAIL)
+          if (i < hostname.length - 1) { 
+            // Write the cookie on the current computed top level domain
+            document.cookie = `${cookie}${Constant.Semicolon}${Constant.Domain}${rootDomain}`;
+            // Once written, check if the cookie was set successfully
+            // If yes, no more action required and return from the function. rootDomain is already set
+            // If no, then continue with the for loop
+            if (document.cookie.indexOf(key) >= 0) { return; }
```
toby-walker, line 208: This might match if the key is a subpart of soe other cookie. Would be cleaner to write a set/getcookie routine somewhere and call those here.

packages/clarity-js/types/data.d.ts
```diff
@@ -138,6 +138,7 @@ export const enum BooleanFlag {
 export const enum Setting {
     Expire = 365, // 1 Year
     SessionExpire = 1, // 1 Day
+    CookieVersion = 1, // Increment this version every time there's a schema change
```
toby-walker, line 141: cookie schema

toby-walker (approved):
:shipit:

packages/clarity-js/src/data/metadata.ts
sarveshnagpal: Good point, done.

packages/clarity-js/src/data/metadata.ts
sarveshnagpal: If cookies are not enabled then we wouldn't execute logic inside the if block and directly return a shortid. It will essentially be same as before - generating a new id.

packages/clarity-js/src/data/metadata.ts
sarveshnagpal: Good point. We already have getCookie and using that instead. Less code + safer.

swaathee (approved):
:shipit:

PR #160, opened 2021-08-04
Prevent running multiple instances of Clarity simultaneously

sarveshnagpal:
Set a "v" property on the global clarity variable so we can look for it before running a second instance.

### DIFF
--- a/packages/clarity-js/src/global.ts
+++ b/packages/clarity-js/src/global.ts
@@ -3,12 +3,20 @@ import * as clarity from "@src/clarity";
 // Expose clarity in a browser environment
 // To be efficient about queuing up operations while Clarity is wiring up, we expose clarity.*(args) => clarity(*, args);
 // This allows us to reprocess any calls that we missed once Clarity is available on the page
+// Once Clarity script bundle is loaded on the page, we also initialize a "v" property that holds current version
+// We use the presence or absence of "v" to determine if we are attempting to run a duplicate instance
 (function(): void {
     if (typeof window !== "undefined") {
         const w = window as any;
-        if (w.clarity && !w.clarity.q) { console.warn("Error CL001: Multiple Clarity tags detected."); }
-        const queue = w.clarity ? (w.clarity.q || []) : [];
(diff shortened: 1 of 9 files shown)


swaathee (approved):
:shipit:

packages/clarity-js/src/global.ts
```diff
@@ -3,12 +3,20 @@ import * as clarity from "@src/clarity";
...
     if (typeof window !== "undefined") {
         const w = window as any;
-        if (w.clarity && !w.clarity.q) { console.warn("Error CL001: Multiple Clarity tags detected."); }
-        const queue = w.clarity ? (w.clarity.q || []) : [];
-        w.clarity = function(method: string, ...args: any[]): void { return clarity[method](...args); }
-        while (queue.length > 0) { w.clarity(...queue.shift()); }
+        const c = 'clarity';
+        
+        // Do not execute or reset global "clarity" variable if a version of Clarity is already running on the page
+        if (w[c] && w[c].v) { return console.warn("Error CL001: Multiple Clarity tags detected."); }
```
toby-walker, line 14: Can we instrument this with a custom variable or event?

packages/clarity-js/src/global.ts
```diff
@@ -3,12 +3,20 @@ import * as clarity from "@src/clarity";
...
 (function(): void {
     if (typeof window !== "undefined") {
         const w = window as any;
-        if (w.clarity && !w.clarity.q) { console.warn("Error CL001: Multiple Clarity tags detected."); }
-        const queue = w.clarity ? (w.clarity.q || []) : [];
-        w.clarity = function(method: string, ...args: any[]): void { return clarity[method](...args); }
-        while (queue.length > 0) { w.clarity(...queue.shift()); }
+        const c = 'clarity';
+        
+        // Do not execute or reset global "clarity" variable if a version of Clarity is already running on the page
```
toby-walker, line 13: Can we track the project id that is also running and then figure out which project id is the source of the duplication?

PR #161, opened 2021-08-05
Persisting server location as part of session information

sarveshnagpal:
- Extension session cookie to store server information and make it forward and backward compatible for future.
- Remove backward compatibility code from 0.6.11 that is no longer needed

### DIFF
--- a/packages/clarity-js/src/data/metadata.ts
+++ b/packages/clarity-js/src/data/metadata.ts
@@ -25,18 +25,10 @@ export function start(): void {
     pageNum: s.count
   }
 
-  // For backward compatibility; remove in future iterations (v0.6.11)
-  if (config.upload && typeof config.upload === Constant.String && (config.upload as string).indexOf(Constant.HTTPS) === 0) {
-    let url = config.upload as string;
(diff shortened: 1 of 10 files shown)


packages/clarity-js/src/data/metadata.ts
```diff
@@ -25,18 +25,10 @@ export function start(): void {
     pageNum: s.count
   }
 
-  // For backward compatibility; remove in future iterations (v0.6.11)
-  if (config.upload && typeof config.upload === Constant.String && (config.upload as string).indexOf(Constant.HTTPS) === 0) {
-    let url = config.upload as string;
```
toby-walker, line 30: Just checking that use of this old version is negligible before you remove this.

toby-walker (approved):
:shipit:

swaathee (approved):
:shipit:

packages/clarity-js/src/data/metadata.ts
sarveshnagpal: Yes confirmed. Thanks for pointing out.

PR #163, opened 2021-08-11
Add support for a fallback upload URL

sarveshnagpal:
- Add support for a fallback upload URL
- Instrument CSP failures
- Add backward / forward compatibility code for simplifying session cookie in future versions
- Renamed log module to explicitly suggest that it's used for internal logging - no change in schema
- Save metadata only after first successful upload to improve data quality


packages/clarity-js/src/core/task.ts
```diff
@@ -74,7 +74,7 @@ function run(): void {
         }).catch((error: Error): void => {
             // If one of the scheduled tasks failed, log, recover and continue processing rest of the tasks
             if (entry.id !== metadata.id()) { return; }
-            log.log(Code.RunTask, error, Severity.Warning);
+            if (error) { internal.log(Code.RunTask, Severity.Warning, error.name, error.message, error.stack); }
```
toby-walker, line 77: When could this be "falsey"?

packages/clarity-js/src/diagnostic/internal.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    encode(Event.Log);
+}
+
+function csp(e: SecurityPolicyViolationEvent): void {
+    let upload = config.upload as string;
+    let parts = upload ? upload.substr(0, upload.indexOf("/", Constant.HTTPS.length)).split(Constant.Dot) : []; // Look for first "/" starting after initial "https://" string
+    let domain = parts.length >= 2 ? parts.splice(-2).join(Constant.Dot) : null;
+    if (domain && e.blockedURI && e.blockedURI.indexOf(domain) >= 0) {
+        log(Code.CSP, Severity.Warning, e.blockedURI, `${e["disposition"]}`);
```
toby-walker, line 34: Should this go reporting endpoint instead?

packages/clarity-js/types/data.d.ts
```diff
@@ -115,7 +115,8 @@ export const enum Code {
     PerformanceObserver = 3,
     CallStackDepth = 4,
     Selector = 5,
-    Metric = 6
+    Metric = 6,
+    CSP = 7
```
toby-walker, line 119: Nit: For acronyms what's the casing rule? Is it CSPRule or CspRule? If former this is fine, if latter should be Csp.

toby-walker (approved):
:shipit:

packages/clarity-js/src/core/task.ts
sarveshnagpal: It's hard to predict one way or another but I kept in parity with older code - which was checking for null error - for now to reduce surprises.

packages/clarity-js/src/diagnostic/internal.ts
sarveshnagpal: I think it should go to log since this gets associated properly with the current page and session. Where as report endpoint logs will have no association. We can revisit this in future versions as we understand more about limitations vs. advantages.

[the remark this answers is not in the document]
sarveshnagpal: We can't. Chances are we won't get to execute this script too if CSP blocks the domain hosting clarity-js.

packages/clarity-js/types/data.d.ts
sarveshnagpal: Changed it to ContentSecurityPolicy for now.

packages/clarity-js/src/data/metadata.ts
```diff
@@ -108,10 +98,19 @@ function tab(): string {
 export function save(): void {
   let ts = Math.round(Date.now());
   let upgrade = config.lean ? BooleanFlag.False : BooleanFlag.True;
-  let upload = typeof config.upload === Constant.String ? config.upload : Constant.Empty;
-  let server = config.server ? config.server.replace(Constant.HTTPS, Constant.Empty) : Constant.Empty;
+  let upload = config.upload && typeof config.upload === Constant.String ? config.upload as string : Constant.Empty;
+  let host: string = Constant.Empty;
+  let path: string = Constant.Empty;
+
+  // The code below checks if the "upload" value is a string, and if so, break it into "server" and "upload" before writing to session storage
```
swaathee, line 105: >"server" and "upload" [](http://example.com/codeflow?start=87&length=21)

nit : host and path

sarveshnagpal: Thanks, updated the comment.

packages/clarity-js/src/data/upload.ts
```diff
@@ -175,51 +177,62 @@ function send(payload: string, zipped: Uint8Array, sequence: number, beacon: boo
...
             // Send back an event only if we were not successful in our first attempt
             if (transitData.attempts > 1) { encode(Event.Upload); }
             // Handle response if it was a 200 response with a valid body
             if (xhr.status === 200 && xhr.responseText) { response(xhr.responseText); }
-            // If we exhausted our retries then trigger Clarity shutdown for this page.
-            // The only exception is if browser decided to not even put the request on the network (status code: 0)
-            if (transitData.attempts > Setting.RetryLimit && xhr.status !== 0) { limit.trigger(Check.Retry); }
+            // If we exhausted our retries then trigger Clarity's shutdown for this page since the data will be incomplete
+            if (xhr.status === 0) { 
+                // And, right before we terminate the session, we will attempt one last time to see if we can use
```
swaathee, line 213: is this code reachable? 0 would always go to the above if right?

swaathee (approved):
:shipit:

packages/clarity-js/src/core/report.ts
```diff
@@ -8,15 +9,18 @@ export function reset(): void {
...
-export function report(message: string): void {
+export function report(check: Check, message: string = null): void {
     // Do not report the same message twice for the same page
     if (history && history.indexOf(message) === -1) {
         const url = config.report;
         if (url && url.length > 0) {
-            let payload = JSON.stringify({m: message, p: data.projectId, u: data.userId, s: data.sessionId, n: data.pageNum} as Report);
+            let payload: Report = {c: check, p: data.projectId, u: data.userId, s: data.sessionId, n: data.pageNum };
+            if (message) payload.m = message;
+            // Using POST request instead of a GET request (img-src) to not violate existing CSP rules
```
toby-walker, line 19: >instead [](http://example.com/codeflow?start=34&length=7)

Can you explain how this will not violate existing CSP rules and why this is better for this?How do we avoid the reporting endpoint itself getting CSP issues?

sarveshnagpal: This was mostly a note for future reference. There is no way to ensure report end-point will comply with CSP rules because third party sites can choose to have their own CSP rules. However, what this does is that it ensures that report end-point stays within the same boundaries that are necessary for core functionality of Clarity. 

More concretely, if a site intentionally adds Clarity from say xyz.com domain, they will likely also add following CSP rule: connect-src: xyz.com (to allow for data to be uploaded to xyz.com using XHR). Which is what report end-point uses right now. If we instead use an image to create a GET request then connect-src permissions will no longer be sufficient and the site will also need to add img-src: xyz.com to CSP before report end-point can work.

packages/clarity-js/src/data/metadata.ts
```diff
@@ -108,18 +100,33 @@ function tab(): string {
...
 
-function track(): void {
-  setCookie(Constant.CookieKey, `${data.userId}${Constant.Pipe}${Setting.CookieVersion}`, Setting.Expire);
+function track(expiry: number): void {
+  // Convert time precision into days to reduce number of bytes we have to write in a cookie
+  // E.g. 1628735962643 / (24*60*60*1000) => 18851 (days) => ejn in base36 (13 bytes => 3 bytes)
+  let end = Math.floor((Date.now() + (Setting.Expire * Time.Day))/Time.Day);
+  // To avoid cookie churn, write user id cookie only once every day
+  if (expiry === null || Math.abs(end - expiry) >= Setting.CookieInterval) {
+    setCookie(Constant.CookieKey, [data.userId, Setting.CookieVersion, end.toString(36)].join(Constant.Pipe), Setting.Expire);
```
toby-walker, line 128: >CookieVersion [](http://example.com/codeflow?start=56&length=13)

Wouldn't this break the cookie decoding for data collection. Should we increment the cookie version?

packages/clarity-js/src/data/metadata.ts
```diff
@@ -108,18 +100,33 @@ function tab(): string {
...
 function supported(target: Window | Document, api: string): boolean {
   try { return !!target[api]; } catch { return false; }
 }
 
-function track(): void {
-  setCookie(Constant.CookieKey, `${data.userId}${Constant.Pipe}${Setting.CookieVersion}`, Setting.Expire);
+function track(expiry: number): void {
+  // Convert time precision into days to reduce number of bytes we have to write in a cookie
+  // E.g. 1628735962643 / (24*60*60*1000) => 18851 (days) => ejn in base36 (13 bytes => 3 bytes)
+  let end = Math.floor((Date.now() + (Setting.Expire * Time.Day))/Time.Day);
```
toby-walker, line 125: >floor [](http://example.com/codeflow?start=17&length=5)

Probably should be ceil to round up, not down?

packages/clarity-js/src/data/metadata.ts
```diff
@@ -151,9 +158,15 @@ function num(string: string, base: number = 10): number {
...
-function user(): string {
+function user(): User {
+  let output: User = { id: shortid(), expiry: null };
   let cookie = getCookie(Constant.CookieKey);
   if(cookie && cookie.length > 0) {
+    // Presence of a valid first-party user cookie implies one of two things below:
+    //   - The project has tracking enabled
+    //   - We received prior consent for this user to enable tracking 
+    // In both cases we can assume tracking is enabled so we can attribute different page views to the same session.
+    config.track = true;
```
toby-walker, line 169: Discuss briefly in a chat.

packages/clarity-js/src/data/metadata.ts
```diff
@@ -209,10 +224,10 @@ function setCookie(key: string, value: string, time: number): void {
           if (i < hostname.length - 1) { 
             // Write the cookie on the current computed top level domain
             document.cookie = `${cookie}${Constant.Semicolon}${Constant.Domain}${rootDomain}`;
-            // Once written, check if the cookie was set successfully
+            // Once written, check if the cookie was set successfully and the value matches exactly with what we set
             // If yes, no more action is required and we can return from the function since rootDomain cookie is already set
```
toby-walker, line 228: Add more to the comment to explain the problem this important change (checking the value) is fixing with an example. In particular when can the query be set but not return the same value?

packages/clarity-js/src/data/metadata.ts
sarveshnagpal: No it won't - this is a first party cookie and available on client only. I was planning to upgrade version for breaking changes only.

packages/clarity-js/src/data/metadata.ts
sarveshnagpal: Sure will do

packages/clarity-js/src/data/metadata.ts
sarveshnagpal: Sure will do although I am not sure if it matters when computing delta.

PR #171, opened 2021-09-16
bug fix: heatmap issues, custom variables

swaathee:
Fixes the following issues 
1. Heatmap does not work for elements inside shadow dom
2. Scroll map height is not set properly 
3. Scrollmap canvas crashes for pages with a very large height 
4.  Custom variables are overriden in some cases and appended in others
5. Add Json+LD for article headlines and creators

### DIFF
--- a/packages/clarity-js/src/layout/node.ts
+++ b/packages/clarity-js/src/layout/node.ts
@@ -1,7 +1,8 @@
 import { Constant, Source } from "@clarity-types/layout";
-import { Code, Severity } from "@clarity-types/data";
+import { Code, Dimension, Severity } from "@clarity-types/data";
 import config from "@src/core/config";
 import * as dom from "./dom";
+import * as dimension from "@src/data/dimension";
 import * as internal from "@src/diagnostic/internal";
 import * as interaction from "@src/interaction";
 import * as mutation from "@src/layout/mutation";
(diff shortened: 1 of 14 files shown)


toby-walker (approved):
:shipit:

packages/clarity-js/src/layout/node.ts
```diff
@@ -104,6 +105,20 @@ export default function (node: Node, source: Source): Node {
                     break;
                 case "NOSCRIPT":
                 case "META":
```
sarveshnagpal, line 107: Let's talk offline about this - since this is a lot of extra code we just introduced.

sarveshnagpal (changes requested):
Let's talk about OG offline. Thanks.

packages/clarity-js/src/layout/node.ts
sarveshnagpal: Synced offline - will revisit this in future iterations.

packages/clarity-js/src/layout/node.ts
```diff
@@ -104,6 +105,20 @@ export default function (node: Node, source: Source): Node {
...
+                    if (Constant.Property in attributes && Constant.Content in attributes) {
+                        var content = attributes[Constant.Content]
```
sarveshnagpal, line 109: We don't use var anymore. Use let for now. We will make a pass and clean everything else.

packages/clarity-js/src/layout/node.ts
```diff
@@ -104,6 +105,20 @@ export default function (node: Node, source: Source): Node {
                     break;
                 case "NOSCRIPT":
```
sarveshnagpal, line 106: Add break here so we don't go through NOSCRIPT tag

sarveshnagpal (approved):

PR #173, opened 2021-09-23
region changes and heatmap bug fixes

swaathee:


packages/clarity-decode/src/layout.ts
```diff
@@ -25,8 +25,9 @@ export function decode(tokens: Data.Token[]): LayoutEvent {
             for (let i = 2; i < tokens.length; i += 3) {
                 let region: Layout.RegionData = {
                     id: tokens[i] as number,
-                    state: tokens[i + 1] as number,
-                    name: tokens[i + 2] as string
+                    interactionState: tokens[i + 1] as number,
```
toby-walker, line 28: Does this beak backwards compat. Could you put the new field at the end to keep it compatible?

packages/clarity-decode/types/layout.d.ts
```diff
@@ -12,6 +12,7 @@ export interface LayoutEvent extends PartialEvent {
 /* Redeclare enums */
 export import Constant = Layout.Constant;
 export import Interaction = Layout.InteractionState;
+export import RegionVisibility = Layout.RegionVisibilityState;
```
toby-walker, line 15: Shouldn't keep name same?

packages/clarity-js/src/layout/node.ts
```diff
@@ -106,9 +106,10 @@ export default function (node: Node, source: Source): Node {
                 case "NOSCRIPT":
                     break;
                 case "META":
-                    if (Constant.Property in attributes && Constant.Content in attributes) {
+                    var key = Constant.Property in attributes ? Constant.Property : Constant.Name in attributes ? Constant.Name : null;
```
toby-walker, line 109: Add bracket to clarify intend meaning of the two ?: expressions.

packages/clarity-js/src/layout/region.ts
```diff
@@ -91,27 +91,30 @@ function handler(entries: IntersectionObserverEntry[]): void {
...
             let id = target ? dom.getId(target) : null;
-            let data = id in regions ? regions[id] : { id, name: regionMap.get(target), state: InteractionState.Rendered };
+            let data = id in regions ? regions[id] : { id, name: regionMap.get(target), interactionState: InteractionState.None, visibilityState: RegionVisibilityState.Rendered };
             
             // For regions that have relatively smaller area, we look at intersection ratio and see the overlap relative to element's area
             // However, for larger regions, area of regions could be bigger than viewport and therefore comparison is relative to visible area
             let viewportRatio = overlap ? (overlap.width * overlap.height * 1.0) / (viewport.width * viewport.height) : 0;
             let visible = viewportRatio > Setting.ViewportIntersectionRatio || entry.intersectionRatio > Setting.IntersectionRatio;
-
+            // If an element is either visible or was visible and has been scorlled to the end
```
toby-walker, line 100: >scorlled [](http://example.com/codeflow?start=75&length=8)

Typo for scrolled.

packages/clarity-js/src/layout/region.ts
```diff
@@ -91,27 +91,30 @@ function handler(entries: IntersectionObserverEntry[]): void {
...
+            let scrolledToEnd = (visible || data.visibilityState == RegionVisibilityState.Visible) && Math.abs(rect.top) + viewport.height > rect.height;
```
toby-walker, line 101: >viewport [](http://example.com/codeflow?start=123&length=8)

Why Math.abs()?

packages/clarity-js/src/layout/region.ts
```diff
@@ -91,27 +91,30 @@ function handler(entries: IntersectionObserverEntry[]): void {
...
             // Process updates to this region, if applicable
-            process(target, data, visible ? InteractionState.Visible : InteractionState.Rendered);
+            process(target, data, data.interactionState, 
+                scrolledToEnd ? RegionVisibilityState.ScrolledToEnd : visible ? RegionVisibilityState.Visible : RegionVisibilityState.Rendered);
```
toby-walker, line 104: Bracked the ?: to make intended grouping clear.

packages/clarity-js/src/layout/region.ts
toby-walker, line 101: Don't understand logic of rect.top + viewport.heigh > rect.height. Can you clarify the logic?

packages/clarity-js/types/layout.d.ts
```diff
@@ -12,12 +12,17 @@ export const enum Source {
 }
 
 export const enum InteractionState {
-    Rendered = 0,
-    Visible = 10,
+    None = 20,
     Clicked = 20,
```
toby-walker, line 16: None and Clicked have same value.

packages/clarity-visualize/src/data.ts
```diff
@@ -52,7 +52,8 @@ export function metric(event: Data.MetricEvent): void {
         // Append region information to metadata
         for (let name in regions) {
             let r = regions[name];
-            let className = r === Layout.Interaction.Visible ? "visible" : (r === Layout.Interaction.Clicked ? "clicked" : Data.Constant.Empty);
+            let className = r.interactionState === Layout.Interaction.Clicked ? "clicked" : Data.Constant.Empty;
+            className += r.visibilityState === Layout.RegionVisibility.ScrolledToEnd ? "visible" : Data.Constant.Empty;
```
toby-walker, line 56: Or Visible state as well? 
Shouldn't the scrolled to end get a separate class?
Also should input interactions get a separate class?

packages/clarity-visualize/src/layout.ts
```diff
@@ -338,6 +338,14 @@ function setAttributes(node: HTMLElement, data: Layout.DomData): void {
 
     // Add an empty ALT tag on all IMG elements
     if (tag === Constant.ImageTag && !node.hasAttribute(Constant.AltAttribute)) { node.setAttribute(Constant.AltAttribute, Constant.Empty); }
+
+    // Prevent auto fill from working on forms during visualize
```
toby-walker, line 342: Thanks for fixing this. Can you add a big more about why this is needed (the fact that it looks like the viewer's own auto-complete information is shown in the playback)?

toby-walker (changes requested):
:clock1:

packages/clarity-visualize/src/data.ts
swaathee: This is mainly for visualization in devtools where we can easily see if a region was scrolled or not - adding too many classes will just confuse us when interpreting. so keeping it minimal with just visible and clicked. For the rest we can always look at the data.

packages/clarity-decode/src/layout.ts
swaathee: Because this is decoded as an array of tokens - so adding to end will not help us differentiate whether it was the id of the next region; I am keeping it in between and using the type to check how to decode this.

packages/clarity-visualize/src/layout.ts
```diff
@@ -338,6 +338,15 @@ function setAttributes(node: HTMLElement, data: Layout.DomData): void {
 
     // Add an empty ALT tag on all IMG elements
     if (tag === Constant.ImageTag && !node.hasAttribute(Constant.AltAttribute)) { node.setAttribute(Constant.AltAttribute, Constant.Empty); }
+
+    // Prevent auto fill from working on forms during visualize. 
+    // During visualization, form fields are filled with saved details from the browser of user who is seeing the visualization
```
toby-walker, line 343: Wording. Make it clear that autocomplete=off disables use of the saved information to autocomplete the form. It sounds like the opposite from this.

packages/clarity-visualize/src/layout.ts
```diff
@@ -338,6 +338,15 @@ function setAttributes(node: HTMLElement, data: Layout.DomData): void {
...
+    if (tag === Constant.FormTag) { 
+        if (node.hasAttribute(Constant.AutoComplte)) {
+            node.removeAttribute(Constant.AutoComplte);
+        }
+        node.setAttribute(Constant.AutoComplte, Constant.NewPassword);
```
toby-walker, line 348: >AutoComplte [](http://example.com/codeflow?start=35&length=11)

Typo for complete.

packages/clarity-js/types/layout.d.ts
```diff
@@ -12,12 +12,17 @@ export const enum Source {
...
-    Visible = 10,
+    None = 16,
     Clicked = 20,
     Input = 30
 }
 
+export const enum RegionVisibility {
+    Rendered = 0,
+    Visible = 10,
+    ScrolledToEnd = 13
```
toby-walker, line 23: Typo or you want an unlucky number by design? :-)

swaathee: By design - trying to have some wiggle room between interaction state and visibility state right now - since we use it for backward compatibility

toby-walker (approved):
:shipit:

alankritj (approved):

PR #174, opened 2021-09-30
Simplify selector logic in clarity-js

sarveshnagpal:
- Stop sending raw selector as part of clarity-js
- Move selector and hash enrichment from clarity-decode to clarity-visualize
- Performance improvement by reducing bytes sent over the wire.
- Enable easier updates to selector logic by simplifying selector handling in clarity-js

### DIFF
--- a/packages/clarity-visualize/src/clarity.ts
+++ b/packages/clarity-visualize/src/clarity.ts
@@ -1,6 +1,7 @@
 import { Visualize } from "@clarity-types/index";
 import { Activity, Constant, MergedPayload, Options, PlaybackState, ScrollMapInfo } from "@clarity-types/visualize";
 import { Data, Interaction, Layout } from "clarity-decode";
+import { helper } from "clarity-js";
 import * as data from "./data";
 import * as heatmap from "./heatmap";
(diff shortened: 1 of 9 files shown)


packages/clarity-visualize/src/clarity.ts
```diff
@@ -58,21 +59,38 @@ export function scrollmap(scrollData: ScrollMapInfo[], avgFold: number, addMarke
 
 export function merge(decoded: Data.DecodedPayload[]): MergedPayload {
     let merged: MergedPayload = { timestamp: null, envelope: null, dom: null, events: [] };
+    let selectors: { [key: number]: string } = {};
+    // Re-arrange decoded payloads in the order of their start time
+    decoded = decoded.sort(sortPayloads);
```
toby-walker, line 64: Should sort by sequence number, not timestamp.

xugao:
jFYI -  I initially made this change #178 to fallback using selector instead of hash to look up element. But with your change, it won't work. I will close my PR for now and I can think of other ways to solve my case (most likely i will build the hashmap for the DOM I am working with)

PR #180, opened 2021-10-13
Bug fixes

sarveshnagpal:
- Fix logic to resume from suspension
- Fix build by renaming visualizer file


xugao:
Thanks for fixing! we should add build check to PRs :)

PR #186, opened 2021-10-30
Minor data updates

sarveshnagpal:
- Support for submit and clipboard actions
- Removing support for connection information and image errors
- Supporting dimensions via config
- Added couple extra metrics to detect automation and event count


packages/clarity-js/src/interaction/submit.ts
```diff
@@ -0,0 +1,30 @@
...
+export function start(): void {
+    reset();
+}
+
+export function observe(root: Node): void {
+    bind(root, "submit", recompute, true);
+}
+
+function recompute(evt: UIEvent): void {
+    state.push({ time: time(), event: Event.Submit, data: { target: target(evt), hash: null } });
```
toby-walker, line 20: Is this correct? The event seems to use a different property Submitterr) andis not a UI event. 

Based on:

[https://developer.mozilla.org/en-US/docs/Web/API/SubmitEvent](https://developer.mozilla.org/en-US/docs/Web/API/SubmitEvent)
[https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/submit_event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/submit_event)
[https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#the-submitevent-interface](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#the-submitevent-interface)

toby-walker, line 20: There is also a "submitter" property on this event. Worth capturing? 

[https://developer.mozilla.org/en-US/docs/Web/API/SubmitEvent](https://developer.mozilla.org/en-US/docs/Web/API/SubmitEvent)

Based on taht I imagine the target points to the form elmeent? Can you confirm?

packages/clarity-js/src/data/metadata.ts
```diff
@@ -41,11 +41,13 @@ export function start(): void {
 
   if (navigator) {
     dimension.log(Dimension.Language, (<any>navigator).userLanguage || navigator.language);
+    metric.max(Metric.Automation, navigator.webdriver ? BooleanFlag.True : BooleanFlag.False);
```
toby-walker, line 44: There's a bunch of stuff for detection automation/headless browser, etc. Is that the goal of this or is it more restricted in inattention?

toby-walker (approved):
:shipit:

PR #188, opened 2021-11-09
Bug fixes and monitoring updates

sarveshnagpal:
- Changing payload for report endpoint to make logging more actionable
- Update beta selector logic to improve data quality
- Bumping version to 0.6.28

### DIFF
--- a/packages/clarity-js/src/core/measure.ts
+++ b/packages/clarity-js/src/core/measure.ts
@@ -1,12 +1,13 @@
 import { Setting } from "@clarity-types/core";
 import { Metric } from "@clarity-types/data";
+import { report } from "@src/core/report";
 import * as metric from "@src/data/metric";
 
 // tslint:disable-next-line: ban-types
 export default function (method: Function): Function {
     return function (): void {
         let start = performance.now();
-        method.apply(this, arguments);
+        try { method.apply(this, arguments); } catch (ex) { throw report(ex); }
         let duration = performance.now() - start;
         metric.sum(Metric.TotalCost, duration);
         if (duration > Setting.LongTask) {
--- a/packages/clarity-js/src/data/metadata.ts
+++ b/packages/clarity-js/src/data/metadata.ts
@@ -89,12 +89,12 @@ export function stop(): void {
   data = null;
 }
 
-export function metadata(cb: MetadataCallback, onupgrade: boolean = true): void {
-  if (data && onupgrade === false) {
+export function metadata(cb: MetadataCallback, wait: boolean = true): void {
+  if (data && wait === false) {
     // Immediately invoke the callback if the caller explicitly doesn't want to wait for the upgrade confirmation
     cb(data, !config.lean);
   } else {
-    // Save the callback for future reference; so we can inform the caller when page gets upgraded
+    // Save the callback for future reference; so we can inform the caller when page gets upgraded and we have a valid playback flag
     callback = cb;
   }
 }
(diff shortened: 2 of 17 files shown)


packages/clarity-js/src/layout/selector.ts
```diff
@@ -24,12 +24,15 @@ export default function(input: SelectorInput, beta: boolean = false): string {
             let selector = `${prefix}${input.tag}${suffix}`;
             let classes = Constant.Class in a && a[Constant.Class].length > 0 ? a[Constant.Class].trim().split(/\s+/) : null;
             if (beta) {
-                // In beta mode, update selector to use "id" field when available
-                // The only exception is if "id" appears to be an auto generated string token, e.g. guid or a random id
+                // In beta mode, update selector to use "id" field when available. There are two exceptions:
+                // (1) if "id" appears to be an auto generated string token, e.g. guid or a random id containing digits
+                // (2) if "id" appears inside a shadow DOM, in which case we continue to prefix up to shadow DOM to prevent conflicts
+                let shadowStart = prefix.indexOf(Constant.ShadowDomTag);
```
xugao, line 30:
```suggestion
                let shadowStart = prefix.lastIndexOf(Constant.ShadowDomTag);
```

to find the last shadow DOM

xugao: should we do the same for `PolyfillShadowDomTag`? not totally sure if needed

packages/clarity-js/src/core/measure.ts
```diff
@@ -1,12 +1,13 @@
...
 import { Metric } from "@clarity-types/data";
+import { report } from "@src/core/report";
 import * as metric from "@src/data/metric";
 
 // tslint:disable-next-line: ban-types
 export default function (method: Function): Function {
     return function (): void {
         let start = performance.now();
-        method.apply(this, arguments);
+        try { method.apply(this, arguments); } catch (ex) { throw report(ex); }
```
MustafaLeithy, line 10: Should we apply this in other places where we think the errors would be of interest?

xugao (approved):

swaathee (approved):
:shipit:

packages/clarity-js/src/core/measure.ts
sarveshnagpal: Thanks for pointing it out. I have intentionally kept this scoped for now - to not change existing behavior. Majority of code flows through this path for performance measurement. Once we have ensured this scoped change hasn't caused any regression - we can then expand to more code paths and cover core.start and data.start methods.

xugao (approved):

packages/clarity-js/src/data/metadata.ts
```diff
@@ -64,13 +64,39 @@ export function start(): void {
...
   rootDomain = null;
+  data = null;
 }
 
-export function metadata(cb: MetadataCallback): void {
-  callback = cb;
+export function metadata(cb: MetadataCallback, onupgrade: boolean = true): void {
+  if (data && onupgrade === false) {
+    // Immediately invoke the callback if the caller explicitly doesn't want to wait for the upgrade confirmation
+    cb(data, !config.lean);
```
xugao, line 95:
```suggestion
    cb(data, onupgrade);
```

xugao: nit: we should also change https://github.com/microsoft/clarity/blob/ddac23ec718f702e9761642a1fdf3b094ff6eb6c/packages/clarity-js/src/data/metadata.ts#L132 to be `callback(data, true)` for readability

sarveshnagpal: This won't be correct since we want to indicate if a valid playback is available for this particular page or not.

ender336 (approved):
:shipit:

PR #190, opened 2021-11-09
expose project id

xugao:
Adding `window.clarity("projectId")` to return `projectId`.

Use case: for extension to know the projectId a given website is associated with


xugao:
closing. using `metadata` instead

Issue #194, opened 2021-11-15
TypeError: Illegal invocation exception


HoussemDjeghri:
Hi, I noticed that sentry is logging an exception thrown by Clarity in my prod env, it says **TypeError: Illegal invocation**, after some digging, I found that this exception is thrown when calling **sendBeacon** function without binding it to the navigator first. It should be called like in the example below, can you check it, please, Maybe adding a try catch could solve this. Here is the [source ](https://xgwang.me/posts/you-may-not-know-beacon/#it-may-throw-error%2C-be-sure-to-catch) of the example.

// ❌
navigator.sendBeacon('/track', 'data');

// ✅
navigator.sendBeacon.bind(navigator)('/track', 'data'); 

<html><body>
<div class=" css-9bj1j7 e9xr8173"><div id="context-browser" class="css-9rnq0x e11dlz2y2"><div class="css-zjik7 e11dlz2y5"><h3>Browser</h3></div></div><div class="css-bjn8wh e11dlz2y1">

name | Chrome
-- | --
version | 72.0.3626

</div></div>
</body>
</html>

<html><body>
<div id="context-os" class="css-9rnq0x e11dlz2y2"><div class="css-zjik7 e11dlz2y5"><h3>Operating System</h3></div></div><div class="css-bjn8wh e11dlz2y1">

Name | Windows
-- | --
Version | 10

</div>
</body>
</html>

![TypeError_ Illegal invocationtest](https://user-images.githubusercontent.com/35101902/141810204-63898a8d-36d8-43a3-9c33-377c05ee65df.jpg)

sarveshnagpal:
Thanks @HoussemDjeghri - Could you please share some more details on why do we need to bind to navigator before calling "sendBeacon". Thanks.

HoussemDjeghri:
Hi @sarveshnagpal, thanks for merging the fix for that issue, but we noticed that our prod app is still using the `0.6.31` version instead of the latest `0.6.32` that contains the fix. Do we need to change something to load the newest version?

sarveshnagpal:
Thanks @HoussemDjeghri for the fix. It's now merged and deployed.

PR #203, opened 2021-12-10
Refactor the send payload function

HoussemDjeghri:
This PR has 2 main changes: 
- Bind the sendBeacon function to the navigator object before passing the reference.
- Surround the sendBeacon function call with a try catch bloc to catch the url parsing errors.

This PR fixes the issue #194

### DIFF
--- a/packages/clarity-js/src/data/upload.ts
+++ b/packages/clarity-js/src/data/upload.ts
@@ -151,9 +151,14 @@ function send(payload: string, zipped: Uint8Array, sequence: number, beacon: boo
         // However, we don't want to rely on it for every payload, since we have no ability to retry if the upload failed.
         // Also, in case of sendBeacon, we do not have a way to alter HTTP headers and therefore can't send compressed payload
         if (beacon && "sendBeacon" in navigator) {
-            // sendBeacon must be binded to the navigator before passing the reference
-            dispatched = navigator.sendBeacon.bind(navigator)(url, payload);
-            if (dispatched) { done(sequence); }
+            try {
+                // sendBeacon must be binded to the navigator before passing the reference
+                dispatched = navigator.sendBeacon.bind(navigator)(url, payload);
+                if (dispatched) { done(sequence); }
+            } catch (error) {
+                // sendBeacon may throw a "TypeError" exception if the url parsing fails
+                console.error(`Payload upload failed: ${error}`);
+            }
         }
 
         // Before initiating XHR upload, we check if the data has already been uploaded using sendBeacon
--- a/packages/clarity-js/src/data/upload.ts
+++ b/packages/clarity-js/src/data/upload.ts
@@ -42,7 +42,7 @@ export function queue(tokens: Token[], transmit: boolean = true): void {
         let now = time();
         let type = tokens.length > 1 ? tokens[1] : null;
         let event = JSON.stringify(tokens);
-        
+
         switch (type) {
             case Event.Discover:
                 discoverBytes += event.length;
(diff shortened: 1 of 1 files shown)


HoussemDjeghri:
This PR fixes the issue #194

HoussemDjeghri:
@sarveshnagpal could you please review this PR, here is an URL example that sendBeacon couldn't parse and throw a "TypeError" or "Illegal Invocation" exception in our prod app earlier today : `/email-confirmed?result=%7B%22status%22%3Afalse%7D`

packages/clarity-js/src/data/upload.ts
```diff
@@ -151,8 +151,14 @@ function send(payload: string, zipped: Uint8Array, sequence: number, beacon: boo
...
         if (beacon && "sendBeacon" in navigator) {
-            dispatched = navigator.sendBeacon(url, payload);
-            if (dispatched) { done(sequence); }
+            try {
+                // sendBeacon must be binded to the navigator before passing the reference
+                dispatched = navigator.sendBeacon.bind(navigator)(url, payload);
+                if (dispatched) { done(sequence); }
+            } catch (error) {
+                // sendBeacon may throw a "TypeError" exception if the url parsing fails
+                console.error(`Payload upload failed: ${error}`);
```
sarveshnagpal, line 160: We don't log debug statements in the production code for performance reasons.

packages/clarity-js/src/data/upload.ts
```diff
@@ -151,8 +151,14 @@ function send(payload: string, zipped: Uint8Array, sequence: number, beacon: boo
         // However, we don't want to rely on it for every payload, since we have no ability to retry if the upload failed.
         // Also, in case of sendBeacon, we do not have a way to alter HTTP headers and therefore can't send compressed payload
         if (beacon && "sendBeacon" in navigator) {
-            dispatched = navigator.sendBeacon(url, payload);
-            if (dispatched) { done(sequence); }
+            try {
+                // sendBeacon must be binded to the navigator before passing the reference
+                dispatched = navigator.sendBeacon.bind(navigator)(url, payload);
```
sarveshnagpal, line 156: I looked up issue you referenced #194 - do you know if this is a Sentry problem? Looked up MDN: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon and it doesn't require binding to navigator before sending sendBeacon. If you could explain more that will help.

sarveshnagpal (changes requested):

packages/clarity-js/src/data/upload.ts
HoussemDjeghri: Okay @sarveshnagpal, I have removed the `console.error` call and replaced it with a "do nothing" comment

packages/clarity-js/src/data/upload.ts
HoussemDjeghri: @sarveshnagpal I couldn't find more details also on the manner, but this fix is used in popular frameworks like next.js, you can check this [commit](https://github.com/blitz-js/next.js/commit/4bb419b473394285249044c174a8ba736669f98f) from the next.js repo , they had the same issue and fixed like I did.

HoussemDjeghri: @sarveshnagpal here is the [issue](https://github.com/vercel/next.js/issues/23856) that mansions this in the next.js repo

HoussemDjeghri: sentry also had the same issue and fixed it like that in this [PR ](https://github.com/getsentry/sentry-javascript/pull/4008)

HoussemDjeghri:
@sarveshnagpal could you review this PR please, we are still receiving the error in PROD.

sarveshnagpal (approved):

PR #214, opened 2022-01-27
Perf improvement & bug fix to improve Angular SPA

sarveshnagpal:
- Drop CSP logging
- Better handling conflicts with Zone.js to improve Clarity on Angular Single Page Applications
- Bug fix in mutation code for nodes that may no longer exist by the time we get to async processing


MustafaLeithy (approved):
LGTM

PR #223, opened 2022-03-01
Extract event changes

swaathee:
Create a unified extract event


packages/clarity-decode/src/data.ts
```diff
@@ -67,6 +67,13 @@ export function decode(tokens: Data.Token[]): DataEvent {
                 variables[tokens[v++] as string] = typeof tokens[v + 1] == Constant.String ? [tokens[v++] as string] : tokens[v++] as string[];
             }
             return { time, event, data: variables };
+        case Data.Event.Extract:
+            let e = 2; // Start from 3rd index since first two are used for time & event
+            let extract = {};
+            while (e < tokens.length) {
+                extract[tokens[e++] as string | number] = tokens[e++] as string;
+            }
+            return { time, event, data: extract };
```
toby-walker, line 76: How do we now what the extracted data is? Should there be some sort of tag/key for it?

packages/clarity-devtools/src/config.ts
```diff
@@ -1,27 +1,22 @@
...
         ],
-        metrics: [
-            [Data.Metric.CartDiscount, 0, "span[data-checkout-discount-amount-target]", 100], /* 0: DOM Text */
-            [Data.Metric.ProductPrice, 1, "Analytics.product.price", 100], /* 1: Javascript */
-            [Data.Metric.CartTotal, 2, "data-checkout-payment-due-target"], /* 2: DOM Attribute */
-        ],
-        dimensions: [
-            [Data.Dimension.ProductBrand, 0, ".productBrand"], /* 0: DOM Text */
-        ]
+        extract: [0, extractKey, "performance.timing"]
```
toby-walker, line 20: Why is this in scope for this PR?

packages/clarity-js/types/layout.d.ts
```diff
@@ -154,6 +154,7 @@ export interface NodeValue {
     hash: [string, string];
     region: number;
     metadata: NodeMeta;
+    fragment: number;
```
toby-walker, line 157: What does this represent?

packages/clarity-js/types/data.d.ts
```diff
@@ -248,6 +250,9 @@ export const enum Constant {
     Accept = "Accept",
     ClarityGzip = "application/x-clarity-gzip",
     Tilde = "~",
+    ArrayStart = "[",
```
toby-walker, line 253: >ArrayStart [](http://example.com/codeflow?start=4&length=10)

ArrayEnd not needed?

packages/clarity-js/types/core.d.ts
```diff
@@ -32,15 +30,11 @@ export const enum Setting {
...
-export const enum RegionFilter {
-    Url = 0,
-    Javascript = 1
-}
-
-export const enum Extract {
-    Text = 0,
-    Javascript = 1,
-    Attribute = 2
+export const enum Source {
```
toby-walker, line 33: Source is too vague and broad a a name. Can you use ExtractSource or something narrower to clarify the meaning?

packages/clarity-js/src/layout/mutation.ts
```diff
@@ -218,27 +218,27 @@ async function processNodeList(list: NodeList, source: Source, timer: Timer): Pr
...
+export function schedule(node: Node, fragment: boolean = false): Node {
   // Only schedule manual trigger for this node if it's not already in the queue
   if (queue.indexOf(node) < 0) { queue.push(node); }
 
   // Cancel any previous trigger before scheduling a new one.
   // It's common for a webpage to call multiple synchronous "insertRule" / "deleteRule" calls.
   // And in those cases we do not wish to monitor changes multiple times for the same node.
   if (timeout) { clearTimeout(timeout); }
-  timeout = setTimeout(trigger, Setting.LookAhead);
+  timeout = setTimeout(() => { trigger(fragment) }, Setting.LookAhead);
```
toby-walker, line 229: >fragment [](http://example.com/codeflow?start=39&length=8)

What does fragment mean in these fns?

packages/clarity-js/src/layout/dom.ts
```diff
@@ -61,9 +62,7 @@ export function parse(root: ParentNode, init: boolean = false): void {
         // Since mutations may happen on leaf nodes too, e.g. text nodes, which may not support all selector APIs.
         // We ensure that the root note supports querySelectorAll API before executing the code below to identify new regions.
         if ("querySelectorAll" in root) {
-            extract.regions(root, config.regions);
-            extract.metrics(root, config.metrics);
-            extract.dimensions(root, config.dimensions);
+            config.regions.forEach(x => root.querySelectorAll(x[1]).forEach(e => region.observe(e, `${x[0]}`))); // Regions
```
toby-walker, line 65: Minor suggestion> Use array destuctureing for the parameters like [a,b]] =\> and indicate by names what they mean.

toby-walker:
Test cases?

packages/clarity-js/src/layout/dom.ts
```diff
@@ -88,11 +87,13 @@ export function add(node: Node, parent: Node, data: NodeInfo, source: Source): v
     let privacy = config.content ? Privacy.Sensitive : Privacy.Text;
     let parentValue = null;
     let regionId = region.exists(node) ? id : null;
+    let fragmentId = null;
 
     if (parentId >= 0 && values[parentId]) {
         parentValue = values[parentId];
         parentValue.children.push(id);
         regionId = regionId === null ? parentValue.region : regionId;
+        fragmentId = parentValue.fragment;
```
toby-walker, line 96: >fragment [](http://example.com/codeflow?start=33&length=8)

Call this field fragmentId if that's what it is.

packages/clarity-js/src/layout/dom.ts
```diff
@@ -115,10 +116,12 @@ export function add(node: Node, parent: Node, data: NodeInfo, source: Source): v
         selector: null,
         hash: null,
         region: regionId,
-        metadata: { active: true, suspend: false, privacy, position: null, size: null }
+        metadata: { active: true, suspend: false, privacy, position: null, size: null },
+        fragment: fragmentId,
     };
 
     updateSelector(values[id]);
+    if (values[id].fragment && updatedFragments.indexOf(values[id].fragment) === -1) { updatedFragments.push(values[id].fragment); }
```
toby-walker, line 124: >values[id].fragment [](http://example.com/codeflow?start=8&length=19)

Minor: Just call this fragmentId as it's clearer and simpler to understand. It's the same value and unchanged from what your used above.

packages/clarity-js/src/layout/dom.ts
```diff
@@ -176,6 +179,17 @@ export function update(node: Node, parent: Node, data: NodeInfo, source: Source)
 
         // Update selector
         updateSelector(value);
+
+        if (value.fragment) {
```
toby-walker, line 183: Add a bit of comment explaining what this is doing and why.

packages/clarity-js/src/layout/dom.ts
```diff
@@ -296,6 +310,9 @@ function updateSelector(value: NodeValue): void {
     value.selector = [selector(s), selector(s, true)];
     value.hash = value.selector.map(x => x ? hash(x) : null) as [string, string];
     value.hash.forEach(h => hashMap[h] = value.id);
+    if (value.hash.some(h => fragments.indexOf(h) !== -1)) {
```
toby-walker, line 313: Are we using the new or the old hashes for this?

packages/clarity-js/src/layout/dom.ts
```diff
@@ -334,6 +351,26 @@ export function updates(): NodeValue[] {
     return output;
 }
 
+export function getFragments(): {} {
+    let output = {};
+    if (fragments) {
+        fragments.forEach(hash => {
```
toby-walker, line 357: Minor: Just use the null coalescing operators like ?., ?[ and ?( for code like this and saves the if() checks

packages/clarity-js/src/layout/dom.ts
```diff
@@ -334,6 +351,26 @@ export function updates(): NodeValue[] {
     return output;
 }
 
+export function getFragments(): {} {
```
toby-walker, line 354: >getFragments [](http://example.com/codeflow?start=16&length=12)

Minor: matchFragments would be a clearer name to me. get is very generic

packages/clarity-js/src/data/extract.ts
```diff
@@ -0,0 +1,127 @@
...
+    let e = config.extract;
+    for (let i = 0; i < e.length; i+=3) {
+        let source = e[i] as Source;
+        let key = e[i+1] as number;
+        switch (source) {
+            case Source.Javascript: 
+                let variable = e[i+2] as string;
+                variables[key] = variable;
+                break;
+            case Source.Cookie:
```
toby-walker, line 25: Not supported?

packages/clarity-js/src/data/extract.ts
```diff
@@ -0,0 +1,127 @@
...
+export function compute(): void {
+    try {
+        for (let v in variables) {
+            let value = evaluate(variables[v]);
+            if (value) { update(v, value); }
+        }
+
+        for (let s in selectors) {
+            let node = document.querySelector(selectors[s] as string) as HTMLElement;
+            if (node) { update(s, node.innerText); }
```
toby-walker, line 48: Idea only: Support innerHtml as well?

packages/clarity-js/src/data/extract.ts
```diff
@@ -0,0 +1,127 @@
...
+// For instance, for a.b.c, it will first check window["a"]. If it exists, it will recursively look at: window["a"]["b"] and finally,
+// return the value for window["a"]["b"]["c"].
+function evaluate(variable: string, type: string = null, base: Object = window): any {
+    let parts = variable.split(Constant.Dot);
+    let syntax = parts.shift();
+    let [first, array, condition] = parse(syntax);
+    let output;
+    if (base && base[first]) {
+        if (array === -1 && match(base[first], condition)) {
+            output = parts.length > 0 ? evaluate(parts.join(Constant.Dot), type, base[first]) : base[first]
```
toby-walker, line 89: This seems a bit inefficient as it reparses, etc. each iteration through the expression. Can's you just iterate over the parts and update the base?

packages/clarity-js/src/data/extract.ts
```diff
@@ -0,0 +1,127 @@
...
+// The function below takes in a variable name in following format: "a.b.c" and safely evaluates its value in javascript context
+// For instance, for a.b.c, it will first check window["a"]. If it exists, it will recursively look at: window["a"]["b"] and finally,
+// return the value for window["a"]["b"]["c"].
+function evaluate(variable: string, type: string = null, base: Object = window): any {
+    let parts = variable.split(Constant.Dot);
+    let syntax = parts.shift();
+    let [first, array, condition] = parse(syntax);
+    let output;
+    if (base && base[first]) {
+        if (array === -1 && match(base[first], condition)) {
```
toby-walker, line 88: >base[first [](http://example.com/codeflow?start=34&length=10)

Make this a variable and use it to  refer to it. Simpler to understand.

packages/clarity-js/src/data/extract.ts
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    return null;
+}
+
+function parse(variable: string): any {
+    let array = variable.indexOf(Constant.ArrayStart);
+    let conditionStart = variable.indexOf(Constant.ConditionStart);
+    let condition = conditionStart > 0 ? variable.substring(conditionStart + 1, variable.indexOf(Constant.ConditionEnd)) : null;
+    let subPart = array > 0 ? variable.substring(0, array) : conditionStart > 0 ? variable.substring(0, conditionStart) : variable;
```
toby-walker, line 111: Too many nested ?:. Can you add brackets to clarify precedence.

packages/clarity-js/src/data/extract.ts
```diff
@@ -0,0 +1,127 @@
...
+    data = {};
+    updateKeys = [];
+    variables = {};
+    selectors = {};
+}
+
+// The function below takes in a variable name in following format: "a.b.c" and safely evaluates its value in javascript context
+// For instance, for a.b.c, it will first check window["a"]. If it exists, it will recursively look at: window["a"]["b"] and finally,
+// return the value for window["a"]["b"]["c"].
+function evaluate(variable: string, type: string = null, base: Object = window): any {
```
toby-walker, line 82: What does this parameter denote? What kind of values are allowed?

packages/clarity-js/src/data/extract.ts
```diff
@@ -0,0 +1,127 @@
...
+    let array = variable.indexOf(Constant.ArrayStart);
+    let conditionStart = variable.indexOf(Constant.ConditionStart);
+    let condition = conditionStart > 0 ? variable.substring(conditionStart + 1, variable.indexOf(Constant.ConditionEnd)) : null;
+    let subPart = array > 0 ? variable.substring(0, array) : conditionStart > 0 ? variable.substring(0, conditionStart) : variable;
+    return [subPart, array, condition];
+}
+
+function str(input: string): string {
+    // Automatically trim string to max of Setting.DimensionLimit to avoid fetching long strings
+    return input ? JSON.stringify(input).substring(0, Setting.ExtractLimit) : input;
```
toby-walker, line 117: This could result in invalid JSON. Can we mark somehow that this has happened?

toby-walker (approved):
:shipit:

packages/clarity-js/types/layout.d.ts
sarveshnagpal: Similar to how region tracks the parent region id, fragment tracks the parent fragment id so we can do O(1) lookup for matching fragment (if any).

packages/clarity-js/types/core.d.ts
```diff
@@ -3,9 +3,8 @@ import * as Data from "./data";
 type TaskFunction = () => Promise<void>;
 type TaskResolve = () => void;
 type UploadCallback = (data: string) => void;
-type Region = [number /* RegionId */, string /* Query Selector */, RegionFilter? /* Region Filter */, string? /* Filter Text */];
-type Metric = [Data.Metric /* MetricId */, Extract /* Extract Filter */, string /* Match Value */, number? /* Scale Factor */];
-type Dimension = [Data.Dimension /* DimensionId */, Extract /* Extract Filter */, string /* Match Value */];
+type Region = [number /* RegionId */, string /* Query Selector */];
+export type Extract = ExtractSource /* Extraction Source */ | number /* Extract Id */ | string  | string[]/* Hash or Query Selector or String Token */;
```
sarveshnagpal, line 7: >/* [](http://example.com/codeflow?start=106&length=2)

nitpick: add a space character

packages/clarity-js/types/core.d.ts
```diff
@@ -119,8 +125,7 @@ export interface Config {
     mask?: string[];
     unmask?: string[];
     regions?: Region[];
-    metrics?: Metric[];
-    dimensions?: Dimension[];
+    extract?: Extract[];
     cookies?: string[];
```
sarveshnagpal, line 129: >cookies?: string[]; [](http://example.com/codeflow?start=4&length=19)

Should we get rid of this and transfer this to extract module?

packages/clarity-js/src/layout/dom.ts
sarveshnagpal, line 124: >if (values[id].fragment && updatedFragments.indexOf( [](http://example.com/codeflow?start=4&length=52)

Can you do this as part of "track" function - since it's purpose is to see if we should track a node for update; we could track fragment there too. It's also shared between add & update - therefore less code duplication.

packages/clarity-js/src/layout/dom.ts
```diff
@@ -176,6 +179,18 @@ export function update(node: Node, parent: Node, data: NodeInfo, source: Source)
 
         // Update selector
         updateSelector(value);
+
+        // if updated node is a part of fragment and the fragment is not being tracked currently, schedule a mutation on the fragment node
```
sarveshnagpal, line 183: >if updated node is a part of fragment [](http://example.com/codeflow?start=11&length=37)

Would it be possible to do this only for the "root" target node of the mutation and not every child node? If so, we can eliminate the cross-talk between dom & mutation and just schedule fragment node mutation from within mutation module while keeping the update fragment logic simple.

packages/clarity-js/src/layout/mutation.ts
```diff
@@ -218,27 +218,27 @@ async function processNodeList(list: NodeList, source: Source, timer: Timer): Pr
...
 
-function schedule(node: Node): Node {
+export function schedule(node: Node, fragment: boolean = false): Node {
   // Only schedule manual trigger for this node if it's not already in the queue
   if (queue.indexOf(node) < 0) { queue.push(node); }
 
   // Cancel any previous trigger before scheduling a new one.
   // It's common for a webpage to call multiple synchronous "insertRule" / "deleteRule" calls.
   // And in those cases we do not wish to monitor changes multiple times for the same node.
   if (timeout) { clearTimeout(timeout); }
```
sarveshnagpal, line 228: >if (timeout) { clearTimeout(timeout); } [](http://example.com/codeflow?start=2&length=39)

If we get two mutations which are part of different fragments within LookAhead period; wouldn't we drop capturing the first fragment mutations?

packages/clarity-js/src/layout/mutation.ts
```diff
@@ -218,27 +218,27 @@ async function processNodeList(list: NodeList, source: Source, timer: Timer): Pr
   }
 }
 
-function schedule(node: Node): Node {
+export function schedule(node: Node, fragment: boolean = false): Node {
```
sarveshnagpal, line 221: >export [](http://example.com/codeflow?start=0&length=6)

Follow up to comment in dom - is it possible to not export it and call it internally?

packages/clarity-js/src/data/extract.ts
```diff
@@ -0,0 +1,139 @@
...
+export let data: {[key: number]: string } = {};
+export let updateKeys: (number | string)[] = [];
+
+let variables : { [key: number]: Syntax[] } = {};
+let selectors : { [key: number]: string } = {};
+
+export function start(): void {
+    let e = config.extract;
+    if (!e) { return; }
+    for (let i = 0; i < e.length; i+=3) {
```
sarveshnagpal, line 18: >for (let i = 0; i < e.length; i+=3) { [](http://example.com/codeflow?start=4&length=37)

Since this depends on configuration - which could technically come in corrupted format - should we wrap it in try / catch like before to ensure we do not break the script for wrong inputs?

packages/clarity-js/src/data/extract.ts
```diff
@@ -0,0 +1,139 @@
...
+            output = filtered;
+        }
+        
+        return str(output);
+    }
+
+    return null;
+}
+function str(input: string): string {
+    // Automatically trim string to max of Setting.DimensionLimit to avoid fetching long strings
```
sarveshnagpal, line 128: >DimensionLimit [](http://example.com/codeflow?start=51&length=14)

ExtractLimit

packages/clarity-js/src/data/extract.ts
```diff
@@ -0,0 +1,139 @@
...
+            case ExtractSource.Cookie: 
+                /*Todo: Add cookie extract logic*/
+                break;
+            case ExtractSource.Text:
+                let match = e[i+2] as string;
+                selectors[key] = match;
+                break;
+            case ExtractSource.Fragment:
+                let fragments =  e[i+2] as string[];
+                dom.setFragments(fragments);
```
sarveshnagpal, line 35: >dom [](http://example.com/codeflow?start=16&length=3)

Usually we want the dependency to be other way. Where DOM depends on Data. Can we either export fragments here and read it in DOM? We will discuss this offline.

packages/clarity-devtools/src/config.ts
sarveshnagpal, line 20: >extractKey [](http://example.com/codeflow?start=21&length=10)

nitpick: Just use 1 here or use the actual ExtractSource.Javascript enum here.

sarveshnagpal:
> export interface UpgradeData {

Define ExtractData here so schema is clear like other events and then decode can reference that while decoding. 

---
Refers to: packages/clarity-js/types/data.d.ts:370 in 0305f5e. [](commit_id = 0305f5ebc78cecb57fbfd7f9c43e517fe1385a53, deletion_comment = False)

sarveshnagpal (changes requested):
:clock1:

sarveshnagpal:
> export interface UpgradeData {

In case you missed this. Rest looks good.

---
In reply to: [1065150503](https://github.com/microsoft/clarity/pull/223#issuecomment-1065150503) [](http://example.com/codeflow?ancestors=1065150503)

---
Refers to: packages/clarity-js/types/data.d.ts:370 in 0305f5e. [](commit_id = 0305f5ebc78cecb57fbfd7f9c43e517fe1385a53, deletion_comment = False)

sarveshnagpal (approved):
:shipit:

Issue #224, opened 2022-03-04
Clarity can remove other listeners on the History API when it reset


ryandabler:
With the current architecture of Clarity's history module it is possible for Clarity to wipe out other scripts that have hooked the `window.history.pushState` and `.replaceState` functions when it cycles through its stop/restart routine.

Essentially what Clarity does when it starts the history module is store off whatever the current state of those methods are on the History API into variables. It then proceeds to wrap those methods with its own custom logic. But then, when the history module stops it resets the methods back to those original variables it created.

The problem comes in when Clarity is the first script to load on a page. At this point the History API's methods are the default ones. But then if another script (say, Script A) comes along afterwards and wraps those some methods, we have Script A wrapping Clarity wrapping History API.

On the first location change using those methods, everything works fine. _But_ Clarity will uninstall itself due to the location change and, in the process, reset the `pushState` method with the default method. At this point Script A has essentially been uninstalled from the History API and won't be present again. This leads to bad data collected by Script A.

I believe Clarity should stay always stay wrapped once it does the wrapping but abort its custom logic if it's in an "off" state. That way other scripts aren't tampered with.

ClaritySupport:
@ryandabler ,this  was resolved. 

[https://github.com/microsoft/clarity/pull/228/files](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fmicrosoft%2Fclarity%2Fpull%2F228%2Ffiles&data=05%7C01%7CToby.Walker%40microsoft.com%7Ca5cf9a56b3d8493222da08dab81625b5%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C638024702567849140%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=HwJymM%2BGodZbOvg9ShA7jJnRiZjU2wez5Vn3aUae1Es%3D&reserved=0)

PR #228, opened 2022-03-15
Proxy the history API only once

ryandabler:
As said in the related issue, other 3rd party scripts might also proxy the History API. If Clarity completely removes its proxy of these methods, it could wipe out those 3rd parties that proxied the API after Clarity did (since Clarity takes its clean version of the methods when it proxies them--those clean methods won't have already been proxied by the 3rd party scripts).

So, Clarity should proxy these methods once and trigger its logic based on some enabled/disabled flag, but otherwise leaving the History API alone after it loads.

Issue #224

### DIFF
--- a/packages/clarity-js/src/core/history.ts
+++ b/packages/clarity-js/src/core/history.ts
@@ -3,33 +3,43 @@ import * as clarity from "@src/clarity";
 import { bind } from "@src/core/event";
 import * as internal from "@src/diagnostic/internal";
 
-let pushState = null;
-let replaceState = null;
+let pushState = history.pushState;
+let replaceState = history.replaceState;
+let enabled = false;
 let url = null;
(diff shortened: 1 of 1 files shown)


packages/clarity-js/src/core/history.ts
```diff
@@ -3,33 +3,43 @@ import * as clarity from "@src/clarity";
...
+let pushState = history.pushState;
+let replaceState = history.replaceState;
+let enabled = false;
 let url = null;
 let count = 0;
 
+// Add a proxy to history.pushState function
+history.pushState = function(): void {
+    // If Clarity is disabled, bypass its History API proxy
+    if (!enabled) {
```
swaathee, line 15: you can use core.active() for this instead of a seperate enabled.

PR #231, opened 2022-03-18
do not override insertrule, deleterule and attachshadow on stop

swaathee:


packages/clarity-js/src/core/history.ts
```diff
@@ -12,24 +12,30 @@ export function start(): void {
...
-        if (check()) {
+    // Add a proxy to history.pushState function
+    if (pushState === null) { 
+        pushState = history.pushState; 
+        history.pushState = function(): void {
             pushState.apply(this, arguments);
-            compute();
-        }
-    };
+            if (core.active() && check()) {
```
sarveshnagpal, line 22: >core.active() && check() [](http://example.com/codeflow?start=16&length=24)

Why do we need check() here? Can we be active without ever passing check()?

packages/clarity-js/src/core/history.ts
```diff
@@ -12,24 +12,30 @@ export function start(): void {
     url = getCurrentUrl();
     count = 0;
     bind(window, "popstate", compute);
-    
+
     // Add a proxy to history.pushState function
-    if (pushState === null) { pushState = history.pushState; }
-    history.pushState = function(): void {
-        if (check()) {
+    // Add a proxy to history.pushState function
```
sarveshnagpal, line 17: >// Add a proxy to history.pushState function [](http://example.com/codeflow?start=4&length=44)

nitpick: Duplicate comment

packages/clarity-js/src/core/history.ts
```diff
@@ -12,24 +12,30 @@ export function start(): void {
...
+    if (pushState === null) {
```
sarveshnagpal, line 18: >pushState === null [](http://example.com/codeflow?start=8&length=18)

What happens if we download Clarity script twice on the page? Would it cause a weird edge case where we override the definition and then override it again because another instance will reset pushState etc. variables?

swaathee: it wouldnt cause any issues - history.pushstate will not be overriden again

packages/clarity-js/src/data/metadata.ts
```diff
@@ -93,10 +91,9 @@ export function metadata(cb: MetadataCallback, wait: boolean = true): void {
   if (data && wait === false) {
     // Immediately invoke the callback if the caller explicitly doesn't want to wait for the upgrade confirmation
     cb(data, !config.lean);
-  } else {
-    // Save the callback for future reference; so we can inform the caller when page gets upgraded and we have a valid playback flag
-    callback = cb;
   }
+  
+  callback.push({callback: cb, wait: wait });
```
toby-walker, line 96: Wouldn't this invoke the call back twice when you execute the if condition? Shouldn't the if() case return without any queuing?

packages/clarity-js/src/data/metadata.ts
```diff
@@ -8,11 +8,10 @@ import * as metric from "@src/data/metric";
 import { set } from "@src/data/variable";
 
 export let data: Metadata = null;
-export let callback: MetadataCallback = null;
+export let callback: MetadataCallbackOptions[] = [];
```
toby-walker, line 11: >callback [](http://example.com/codeflow?start=11&length=8)

Shouldn't this be named "callbacks" (plural) as it's a list of registered callbacks.

packages/clarity-js/src/data/metadata.ts
toby-walker, line 96: >h [](http://example.com/codeflow?start=14&length=1)

Should we register a callback when the if condition is executed?

toby-walker (changes requested):
:clock1:

toby-walker (approved):
:shipit:

PR #240, opened 2022-04-19
Adding trace event in diagnostics for backend usage

maryamshal:
- Created a trace event to hold data for tracing upload requests
- Adding tagTraceId to correlate the upload requests with their tag request
- tagTraceId will be set through the tag response


packages/clarity-js/types/index.d.ts
```diff
@@ -16,6 +16,7 @@ interface Clarity {
   set: (variable: string, value: string | string[]) => void;
   identify: (userId: string, sessionId?: string, pageId?: string) => void;
   metadata: (callback: Data.MetadataCallback, wait?: boolean) => void;
+  trace: (tagTraceId?: string) => void;
```
swaathee, line 19: I felt this was better as a config than as a seperate function call since this is more for diagnostics than something we want to expose to website owners for setting variables.

PR #241, opened 2022-04-27
Update envelope

Hassanosama:
A new field in envelope is introduced in backend called `ApplicationPlatform`.
This field is mainly to identify the platform of the application which invoked collect API. The useability of such a field comes with introducing new Clarity Platforms, e.g., Clarity for Android Apps, IOS Apps, Desktop Apps and so on.

### DIFF
--- a/packages/clarity-decode/src/data.ts
+++ b/packages/clarity-decode/src/data.ts
@@ -89,6 +89,7 @@ export function envelope(tokens: Data.Token[]): Data.Envelope {
         sessionId: tokens[6] as string,
         pageNum: tokens[7] as number,
         upload: tokens[8] as Data.Upload,
-        end: tokens[9] as Data.BooleanFlag
+        end: tokens[9] as Data.BooleanFlag,
+        applicationPlatform: tokens[10] as Data.ApplicationPlatform
     };
 }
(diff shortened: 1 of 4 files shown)


packages/clarity-decode/src/data.ts
```diff
@@ -89,6 +89,7 @@ export function envelope(tokens: Data.Token[]): Data.Envelope {
         sessionId: tokens[6] as string,
         pageNum: tokens[7] as number,
         upload: tokens[8] as Data.Upload,
-        end: tokens[9] as Data.BooleanFlag
+        end: tokens[9] as Data.BooleanFlag,
+        applicationPlatform: tokens[10] as Data.ApplicationPlatform
```
swaathee, line 93: rename to platform  or app everywhere

packages/clarity-js/src/data/metadata.ts
```diff
@@ -23,7 +23,8 @@ export function start(): void {
     projectId: config.projectId || hash(location.host),
     userId: u.id,
     sessionId: s.session,
-    pageNum: s.count
+    pageNum: s.count,
+    applicationPlatform: ApplicationPlatform.WebApp
```
swaathee, line 27: you don't need this here - remove

PR #263, opened 2022-08-09
Masking logic update + bug fixes

sarveshnagpal:
- Addressing user feedback to adjust default masking capabilities
- Updating unit tests to reflect new logic
- Updating mousewheel -> wheel event for better browser compatibility
- Bumping up version to 0.6.38

### DIFF
--- a/packages/clarity-js/src/core/scrub.ts
+++ b/packages/clarity-js/src/core/scrub.ts
@@ -92,9 +92,9 @@ function redact(value: string): string {
                 if (array === null) { array = value.split(Data.Constant.Empty); }
                 // Work on a token at a time so we don't have to apply regex to a larger string
                 let token = value.substring(spaceIndex + 1, hasWhitespace ? i : i + 1);
-                // Check if unicode regex is supported, otherwise fallback to legacy regex
+                // Check if unicode regex is supported, otherwise fallback to calling mask function on this token
                 if (currencyRegex.unicode) {
-                    // Do not mask if the token contains a currency symbol
+                    // Do not redact information if the token contains a currency symbol
                     token = token.match(currencyRegex) ? token : token.replace(letterRegex, Data.Constant.Letter).replace(digitRegex, Data.Constant.Digit);
                 } else {
                     token = mask(token);
--- a/packages/clarity-js/src/core/scrub.ts
+++ b/packages/clarity-js/src/core/scrub.ts
@@ -2,6 +2,13 @@ import { Privacy } from "@clarity-types/core";
 import * as Data from "@clarity-types/data";
 import * as Layout from "@clarity-types/layout";
 
+// Regular expressions using unicode property escapes
+// Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Unicode_Property_Escapes
+const digitRegex = /\p{N}/gu;
+const letterRegex = /\p{L}/gu;
+const currencyRegex = /\p{Sc}/u;
+const catchallRegex = /\S/gi;
+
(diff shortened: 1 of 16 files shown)


packages/clarity-js/types/data.d.ts
```diff
@@ -247,7 +247,9 @@ export const enum Constant {
     UserId = "userId",
     SessionId = "sessionId",
     PageId = "pageId",
-    Mask = "•",
+    Mask = "•", // Placeholder character for explicitly masked content
+    Digit = "•", // Placeholder character for digits
```
toby-walker, line 251: >• [](http://example.com/codeflow?start=13&length=1)

Could we use a digit instead?

toby-walker:
>         hasWhitespace = c === Data.Character.Tab || c === Data.Character.NewLine || c === Data.Character.Return || c === Data.Character.Blank;

Similarly this wouldn't handle other whitespace character classes -- e.g., Japanese might use a different character for a blank.

---
Refers to: packages/clarity-js/src/core/scrub.ts:86 in 9419b65. [](commit_id = 9419b65a1e6f3d5684cce96173e1d5cd836e79b7, deletion_comment = False)

toby-walker (approved):
:shipit:

packages/clarity-js/types/data.d.ts
sarveshnagpal: I will tackle this in a follow on change. Currently keep it consistent with existing masking expectation.

sarveshnagpal:
>         hasDigit = hasDigit || (c >= Data.Character.Zero && c <= Data.Character.Nine); // Check for digits in the current word

For now yes; since there are performance implications of moving regex outside this block. I will tackle that in a separate change.

---
In reply to: [1210040837](https://github.com/microsoft/clarity/pull/263#issuecomment-1210040837) [](http://example.com/codeflow?ancestors=1210040837,1210040837)

---
Refers to: packages/clarity-js/src/core/scrub.ts:84 in 9419b65. [](commit_id = 9419b65a1e6f3d5684cce96173e1d5cd836e79b7, deletion_comment = False)

sarveshnagpal:
>         hasWhitespace = c === Data.Character.Tab || c === Data.Character.NewLine || c === Data.Character.Return || c === Data.Character.Blank;

Yes known but agreed to defer it to a different change to not regress performance.

---
In reply to: [1210041313](https://github.com/microsoft/clarity/pull/263#issuecomment-1210041313) [](http://example.com/codeflow?ancestors=1210041313,1210041313)

---
Refers to: packages/clarity-js/src/core/scrub.ts:86 in 9419b65. [](commit_id = 9419b65a1e6f3d5684cce96173e1d5cd836e79b7, deletion_comment = False)

packages/clarity-js/src/core/scrub.ts
```diff
@@ -82,7 +90,18 @@ function redact(value: string): string {
             // Performance optimization: Lazy load string -> array conversion only when required
             if (hasDigit || hasEmail) {
                 if (array === null) { array = value.split(Data.Constant.Empty); }
-                mutate(array, spaceIndex, hasWhitespace ? i : i + 1);
+                // Work on a token at a time so we don't have to apply regex to a larger string
+                let token = value.substring(spaceIndex + 1, hasWhitespace ? i : i + 1);
+                // Check if unicode regex is supported, otherwise fallback to legacy regex
+                if (currencyRegex.unicode) {
```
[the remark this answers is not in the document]
sarveshnagpal: Technically redact and mask are different functions. Mask strips off every character while redact only looks for sensitive content through digits and email address. I will update the comment to make the distinction clearer.

ender336 (approved):

PR #270, opened 2022-08-26
Add support for alpha selector

sarveshnagpal:
- Rename stable selector to alpha selector
- Change algorithm of how alpha selector is computed
- Few additional bug fixes

### DIFF
--- a/packages/clarity-js/src/layout/dom.ts
+++ b/packages/clarity-js/src/layout/dom.ts
@@ -239,8 +239,9 @@ function privacy(node: Node, value: NodeValue, parent: NodeValue): void {
         case tag === Constant.TextTag:
             // If it's a text node belonging to a STYLE or TITLE tag or one of scrub exceptions, then capture content
             let pTag = parent && parent.data ? parent.data.tag : Constant.Empty;
-            let pSelector = parent && parent.selector ? parent.selector[Selector.Beta] : Constant.Empty;
-            metadata.privacy = [Constant.StyleTag, Constant.TitleTag, Constant.SvgStyle].includes(pTag) || override.some(x => pSelector.indexOf(x) >= 0) ? Privacy.None : current;
+            let pSelector = parent && parent.selector ? parent.selector[Selector.Default] : Constant.Empty;
+            let tags : string[] = [Constant.StyleTag, Constant.TitleTag, Constant.SvgStyle];
+            metadata.privacy = tags.includes(pTag) || override.some(x => pSelector.indexOf(x) >= 0) ? Privacy.None : current;
             break;
         case Constant.Type in attributes:
             // If this node has an explicit type assigned to it, go through masking rules to determine right privacy setting
@@ -303,6 +304,7 @@ function updateSelector(value: NodeValue): void {
     value.selector = [selector.get(s, Selector.Alpha), selector.get(s, Selector.Beta)];
     value.hash = value.selector.map(x => x ? hash(x) : null) as [string, string];
     value.hash.forEach(h => hashMap[h] = value.id);
+    // Match fragment configuration against both alpha and beta hash
     if (value.hash.some(h => extract.fragments.indexOf(h) !== -1)) {
         value.fragment = value.id;
     }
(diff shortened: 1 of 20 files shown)


packages/clarity-js/src/layout/dom.ts
```diff
@@ -236,10 +237,10 @@ function privacy(node: Node, value: NodeValue, parent: NodeValue): void {
             metadata.privacy = Privacy.Text;
             break;
         case tag === Constant.TextTag:
-            // If it's a text node belonging to a STYLE or TITLE tag or one of SCRUB_EXCEPTIONS, then capture content
+            // If it's a text node belonging to a STYLE or TITLE tag or one of scrub exceptions, then capture content
             let pTag = parent && parent.data ? parent.data.tag : Constant.Empty;
-            let pSelector = parent && parent.selector ? parent.selector[Selector.Stable] : Constant.Empty;
-            metadata.privacy = pTag === Constant.StyleTag || pTag === Constant.TitleTag || override.some(x => pSelector.indexOf(x) >= 0) ? Privacy.None : current; 
+            let pSelector = parent && parent.selector ? parent.selector[Selector.Beta] : Constant.Empty;
```
toby-walker, line 242: Minor suggestion: Define a constant one place that is the production one and use that to refer the one in use. That way flipping beteween them should be a one-line change that ripples to everywhere. Too easy otherwise to miss them buried. Then make sure there is no code anywhere referring to Selector.{Alpha/Beta}

packages/clarity-js/src/layout/dom.ts
```diff
@@ -298,8 +299,8 @@ function updateSelector(value: NodeValue): void {
     let prefix = parent ? parent.selector : null;
     let d = value.data;
     let p = position(parent, value);
-    let s: SelectorInput = { tag: d.tag, prefix, position: p, attributes: d.attributes };
-    value.selector = [selector(s), selector(s, true)];
+    let s: SelectorInput = { id: value.id, tag: d.tag, prefix, position: p, attributes: d.attributes };
+    value.selector = [selector.get(s, Selector.Alpha), selector.get(s, Selector.Beta)];
     value.hash = value.selector.map(x => x ? hash(x) : null) as [string, string];
     value.hash.forEach(h => hashMap[h] = value.id);
     if (value.hash.some(h => extract.fragments.indexOf(h) !== -1)) {
```
toby-walker, line 306: If I read this correctly it matches on either of the two types of hashes. Is that intentional? If so please just add a comment that is what this should do.

packages/clarity-visualize/src/enrich.ts
```diff
@@ -43,21 +44,21 @@ export class EnrichHelper {
...
             /* For backward compatibility, continue populating current selector and hash like before in addition to beta selector and hash */
-            let input: Layout.SelectorInput = { tag: d.tag, prefix: parent ? [parent.stable, parent.beta] : null, position: node.position, attributes };
+            let input: Layout.SelectorInput = { id: d.id, tag: d.tag, prefix: parent ? [parent.alpha, parent.beta] : null, position: node.position, attributes };
 
             // Get stable selector
             // We intentionally use "null" value for empty selectors to keep parity with v0.6.25 and before.
-            let selector = helper.selector(input);
-            d.selector = selector.length > 0 ? selector : null;
-            d.hash = selector.length > 0 ? helper.hash(d.selector) : null;
+            let selectorAlpha = helper.selector.get(input, Layout.Selector.Alpha);
```
toby-walker, line 51: Minor: Wuldn't it be a bit cleaner to return both the selectosr and hashes from this function? Too much code dup calling this function and using its values.

toby-walker (approved):
:shipit:

packages/clarity-js/src/layout/dom.ts
sarveshnagpal: Good point. Fixed it.

packages/clarity-js/src/layout/dom.ts
sarveshnagpal: Sounds good.

packages/clarity-visualize/src/enrich.ts
sarveshnagpal: It's a good idea. I will tackle this in a separate change because it may have impact on external libraries depending on this function.

PR #273, opened 2022-09-06
Fix timeout callbacks with undefined arguments 

moustaphaalaa:
- fixed an issue where in some cases the callback function in the timeout is passed with different or missing arguments. this is caused due to the change of the context. I've created a wrapper for the callback with the arguments directly passed as parameters

### DIFF
--- a/packages/clarity-js/src/core/timeout.ts
+++ b/packages/clarity-js/src/core/timeout.ts
@@ -2,7 +2,7 @@ import { Event } from "@clarity-types/data";
 import measure from "./measure";
 
 export function setTimeout(handler: (event?: Event | boolean) => void, timeout?: number, event?: Event): number {
-    return window.setTimeout(measure(handler), timeout, event);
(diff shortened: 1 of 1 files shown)


packages/clarity-js/src/core/timeout.ts
```diff
@@ -2,7 +2,7 @@ import { Event } from "@clarity-types/data";
 import measure from "./measure";
 
 export function setTimeout(handler: (event?: Event | boolean) => void, timeout?: number, event?: Event): number {
-    return window.setTimeout(measure(handler), timeout, event);
+    return window.setTimeout(() => measure(handler)(event), timeout);
```
sarveshnagpal, line 5: This will change the order of operations and also the current usage is correct per the spec: https://developer.mozilla.org/en-US/docs/Web/API/setTimeout.

sarveshnagpal (changes requested):

PR #276, opened 2022-09-15
adding callback support for server initiated actions

ender336:


packages/clarity-js/types/data.d.ts
```diff
@@ -265,7 +265,8 @@ export const enum Constant {
     Tilde = "~",
     ArrayStart = "[",
     ConditionStart = "{",
-    ConditionEnd = "}"
+    ConditionEnd = "}",
+    Action = "Action"
```
sarveshnagpal, line 269: Let's make this uppercase to stay consistent with other responses.

packages/clarity-js/src/data/upload.ts
```diff
@@ -252,5 +252,8 @@ function response(payload: string): void {
             // Upgrade current session to send back playback information
             clarity.upgrade(Constant.Auto);
             break;
+        case Constant.Action:
+            config.action && config.action(payload.split(" ")[1]);
```
sarveshnagpal, line 256: Since the response comes from a server and there's a potential for a bug -- we should check if the response has a space before referencing [1] to avoid null exceptions.

sarveshnagpal (approved):

PR #283, opened 2022-09-30
exposing a few methods useful for finding elements on page

ender336:


packages/clarity-js/src/diagnostic/fraud.ts
```diff
@@ -10,6 +10,7 @@ export let data: FraudData;
 export function start(): void {
     history = [];
     metric.max(Metric.Automation, navigator.webdriver ? BooleanFlag.True : BooleanFlag.False);
+    metric.max(Metric.Iframed, window.top !== window.self ? BooleanFlag.True : BooleanFlag.False);
```
sarveshnagpal, line 13: We need to wrap this in try / catch or this will throw exception if the the current window doesn't have access to top most window.

packages/clarity-js/src/core/version.ts
```diff
@@ -1,2 +1,2 @@
-let version = "0.6.42";
+let version = "0.6.43";
```
sarveshnagpal, line 1: Please take a look at other PRs; there are few more files where you will need to update the version - including decode, visualize, etc.

packages/clarity-js/src/layout/dom.ts
```diff
@@ -307,6 +307,12 @@ function updateSelector(value: NodeValue): void {
     }
 }
 
+export function hashText(hash: string): string {
+    let id = lookup(hash);
+    let node = getNode(id);
+    return node != null ? node.textContent : '';
```
sarveshnagpal, line 313: Only look at a substring with max length of 20 or something reasonable so we accidentally don't send back the whole text.

packages/clarity-js/types/data.d.ts
```diff
@@ -96,7 +96,8 @@ export const enum Metric {
     Mobile = 27,
     UploadTime = 28,
     SinglePage = 29,
-    UsedMemory = 30
+    UsedMemory = 30,
+    Iframed = 30,
```
sarveshnagpal, line 100: Change this to 31 so it's not conflicting with already defined value.

sarveshnagpal (changes requested):

sarveshnagpal (approved):

PR #307, opened 2022-11-21
pass current url with extractEvent

moustaphaalaa:


packages/clarity-js/src/data/encode.ts
```diff
@@ -112,6 +112,8 @@ export default function(event: Event): void {
                 tokens.push(e);
                 tokens.push(extract.data[e]);
             }
+            tokens.push('u')
```
sarveshnagpal, line 115: We shouldn't hardcode this and include it as part of the interface: [ExtractData](https://github.com/microsoft/clarity/blob/master/packages/clarity-js/types/data.d.ts#L395)

sarveshnagpal (changes requested):

PR #316, opened 2022-12-09
Clarity 0.7.0 Updates

sarveshnagpal:
- Using event timestamp instead of performance.now()
- Strengthening masking by default
- Updating unit tests
- Code refactoring for performance
- Corresponding decode and visualization changes


lerna.json
```diff
@@ -2,7 +2,7 @@
   "packages": [
     "packages/*"
   ],
-  "version": "0.6.43",
+  "version": "0.6.44",
```
toby-walker, line 5: I'd prefer to do this 0.7.0 Not sure why we don't use a wider range of version to indicate how big the change is.

packages/clarity-decode/src/interaction.ts
```diff
@@ -23,7 +25,7 @@ export function decode(tokens: Data.Token[]): InteractionEvent {
         case Data.Event.Click:
             let clickHashes = (tokens[12] as string).split(Data.Constant.Dot);
             let clickData: ClickData = {
-                target: tokens[2] as number,
+                target: Math.abs(tokens[2] as number),
```
toby-walker, line 28: Why the abs? Not obvious so please add comment explaining this.

packages/clarity-decode/src/core.ts
```diff
@@ -0,0 +1,10 @@
+import { Data, Layout } from "clarity-js";
+
+export function expand(input: string): [string, string] {
+    let parts = input.split(Data.Constant.Dot);
```
toby-walker, line 4: Suggestion: Write as [length, checkum] = input.split

Both shorter and simpler.

packages/clarity-decode/src/core.ts
```diff
@@ -0,0 +1,10 @@
+import { Data, Layout } from "clarity-js";
+
+export function expand(input: string): [string, string] {
```
toby-walker, line 3: Maybe demangle would be clearer? Exapnd is very vague and generic.

packages/clarity-decode/types/interaction.d.ts
```diff
@@ -3,13 +3,18 @@ import { PartialEvent } from "./core";
...
+    checksum?: string;
 }
 
 export interface TimelineData extends Interaction.TimelineData {
     hashBeta: string;
 }
 
-export interface InputEvent extends PartialEvent { data: Interaction.InputData; }
+export interface InputData extends Interaction.InputData {
+    checksum?: string;
```
toby-walker, line 14: Why not put this field in the base type?

sarveshnagpal: Because it's part of the value (where possible) and not a separate field.

lerna.json
sarveshnagpal: Good point. Will update to 0.7.0.

packages/clarity-js/src/core/scrub.ts
```diff
@@ -64,11 +84,8 @@ function mask(value: string): string {
 
 function mangleToken(value: string): string {
     let length = ((Math.floor(value.length / Data.Setting.WordLength) + 1) * Data.Setting.WordLength);
-    let output: string = Layout.Constant.Empty;
-    for (let i = 0; i < length; i++) {
-        output += i > 0 && i % Data.Setting.WordLength === 0 ? Data.Constant.Space : Data.Constant.Mask;
-    }
-    return output;
+    let checksum = value.length >= Data.Setting.WordLength && config.fraud ? hash(value, Data.Setting.ChecksumPrecision) : Data.Constant.Zero;
```
toby-walker, line 87: Use a value that is not a legal hash like -1 or empty.

packages/clarity-decode/src/core.ts
sarveshnagpal: Thanks, done.

packages/clarity-decode/src/core.ts
sarveshnagpal: Agree, renamed.

packages/clarity-js/src/core/scrub.ts
sarveshnagpal: Went with empty.

packages/clarity-decode/src/layout.ts
```diff
@@ -125,8 +126,9 @@ function process(node: any[] | number[], tagIndex: number): DomData {
             let k = token.substr(0, keyIndex);
             let v = token.substr(keyIndex + 1);
             attributes[k] = v;
+            if (mangled && k === "value") { [attributes[k], output.checksum] = expand(v); }
         } else if (output.tag === Layout.Constant.TextTag) {
-            value = masked ? unmask(token) : token;
+            value = mangled ? unmask(token) : token;
```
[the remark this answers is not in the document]
sarveshnagpal: One has checksum and other doesn't. Renamed expand => demangle based on your other comment.

packages/clarity-js/src/data/metadata.ts
```diff
@@ -19,32 +20,35 @@ export function start(): void {
   // Populate ids for this page
   let s = session();
   let u = user();
-  data = {
-    projectId: config.projectId || hash(location.host),
-    userId: u.id,
-    sessionId: s.session,
-    pageNum: s.count
-  }
+  let projectId = config.projectId || hash(location.host);
```
toby-walker, line 23: Hashing the host name seems weird. In what cases would we do this?

packages/clarity-visualize/src/interaction.ts
```diff
@@ -34,7 +34,13 @@ export class InteractionHelper {
         let de = doc.documentElement;
         let scrollTarget = this.layout.element(data.target as number) as HTMLElement || doc.body;
         let scrollable = scrollTarget.scrollHeight > scrollTarget.clientHeight || scrollTarget.scrollWidth > scrollTarget.clientWidth;
-        if (scrollTarget && scrollable) { scrollTarget.scrollTo(data.x, data.y); }
+        if (scrollTarget && scrollable) {
+            scrollTarget.scrollTo(data.x, data.y);
```
toby-walker, line 38: Suggestion: What is this doing? Add comment to explain this as not self-evident.

packages/clarity-decode/types/interaction.d.ts
toby-walker: Sorry, don't understand the response. Can you clarify. Why only needed in these two cases?

packages/clarity-decode/types/interaction.d.ts
```diff
@@ -3,13 +3,18 @@ import { PartialEvent } from "./core";
 
 export interface ClickData extends Interaction.ClickData {
     hashBeta: string;
+    checksum?: string;
 }
 
 export interface TimelineData extends Interaction.TimelineData {
     hashBeta: string;
 }
```
toby-walker, line 11: Where are the decode change that consume the checksum field here? Where are the changes that send it?

packages/clarity-js/types/data.d.ts
```diff
@@ -126,7 +130,9 @@ export const enum Dimension {
     Platform = 22,
     PlatformVersion = 23,
     Brand = 24,
-    Model = 25
+    Model = 25,
+    DevicePixelRatio = 26,
```
toby-walker, line 134: This is a number but floating point (and don't want to convert to int). Is that why it's a dimension?

packages/clarity-decode/types/interaction.d.ts
sarveshnagpal: Cleaned it up -- not part of the base type.

packages/clarity-js/src/data/metadata.ts
sarveshnagpal: This is not new code and this is mostly for dev tools extension so it can work against any website as part of the testing.

packages/clarity-visualize/src/interaction.ts
sarveshnagpal: This was a user reported bug. In some cases, because of active CSS, documentElement is not scrollable and therefore we need to fall back to body. I will add comments.

packages/clarity-decode/types/interaction.d.ts
sarveshnagpal: Simplified this. We now support change event.

packages/clarity-js/types/data.d.ts
sarveshnagpal: Yes, right.

packages/clarity-js/types/data.d.ts
```diff
@@ -98,6 +99,10 @@ export const enum Metric {
     SinglePage = 29,
     UsedMemory = 30,
     Iframed = 31,
+    MaxTouchPoints = 32,
+    HardwareConcurrency = 33,
+    DeviceMemory = 34,
+    SaveData = 35
```
toby-walker, line 105: Removed this, not needed.

sarveshnagpal: removed

MustafaLeithy (approved):
LGTM

PR #319, opened 2022-12-15
handling identify, set and consent while clarity is stopped

ender336:


packages/clarity-decode/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clarity-decode",
-  "version": "0.7.1",
+  "version": "2",
```
sarveshnagpal, line 3: Please update this to 0.7.2

sarveshnagpal (changes requested):
We already maintain queue via tag snippet. I will follow up offline to understand this more.

packages/clarity-js/src/data/variable.ts
```diff
@@ -9,14 +9,23 @@ export function start(): void {
...
+        log(Constant.UserId, [userId]);
+        log(Constant.SessionId, [sessionId]);
+        log(Constant.PageId, [pageId]);
+    } else {
+        core.queue("identify", arguments);
+    }
+    
 }
 
 function log(variable: string, value: string[]): void {
```
toby-walker, line 31: Shouldn't we apply similar logic to above here?

ender336: I don't think so - this function isn't exported and the queuing is handler in its callers

PR #343, opened 2023-02-06
Bug fixes based on user feedback

sarveshnagpal:
- Improve logic to determine which elements get suspended from continuous mutations
- Better queuing while Clarity is stopped or suspended
- Visualization improvements to handle third party resources

### DIFF
--- a/packages/clarity-js/src/global.ts
+++ b/packages/clarity-js/src/global.ts
@@ -2,5 +2,5 @@ import * as queue from "@src/queue";
 
 // Process anything that was queued up before the script loaded
 (function(): void {
-    queue.clear();
+    queue.process();
 })();
--- a/packages/clarity-js/src/global.ts
+++ b/packages/clarity-js/src/global.ts
@@ -1,22 +1,6 @@
-import * as clarity from "@src/clarity";
+import * as queue from "@src/queue";
 
-// Expose clarity in a browser environment
-// To be efficient about queuing up operations while Clarity is wiring up, we expose clarity.*(args) => clarity(*, args);
-// This allows us to reprocess any calls that we missed once Clarity is available on the page
-// Once Clarity script bundle is loaded on the page, we also initialize a "v" property that holds current version
-// We use the presence or absence of "v" to determine if we are attempting to run a duplicate instance
+// Process anything that was queued up before the script loaded
 (function(): void {
-    if (typeof window !== "undefined") {
-        const w = window as any;
-        const c = 'clarity';
-        
-        // Do not execute or reset global "clarity" variable if a version of Clarity is already running on the page
-        if (w[c] && w[c].v) { return console.warn("Error CL001: Multiple Clarity tags detected."); }
-        
-        // Re-wire global "clarity" variable to map it to current instance of Clarity
-        const queue = w[c] ? (w[c].q || []) : [];
-        w[c] = function(method: string, ...args: any[]): void { return clarity[method](...args); }
-        w[c].v = clarity.version;
-        while (queue.length > 0) { w[c](...queue.shift()); }
-    }
+    queue.clear();
 })();
--- a/packages/clarity-js/src/layout/mutation.ts
+++ b/packages/clarity-js/src/layout/mutation.ts
@@ -124,12 +124,13 @@ async function process(): Promise<void> {
   task.start(timer);
   while (mutations.length > 0) {
     let record = mutations.shift();
+    let instance = time();
     for (let mutation of record.mutations) {
       let state = task.state(timer);
       if (state === Task.Wait) { state = await task.suspend(timer); }
       if (state === Task.Stop) { break; }      
       let target = mutation.target;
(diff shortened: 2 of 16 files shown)


packages/clarity-js/src/global.ts
```diff
@@ -1,22 +1,6 @@
...
-        // Do not execute or reset global "clarity" variable if a version of Clarity is already running on the page
-        if (w[c] && w[c].v) { return console.warn("Error CL001: Multiple Clarity tags detected."); }
-        
-        // Re-wire global "clarity" variable to map it to current instance of Clarity
-        const queue = w[c] ? (w[c].q || []) : [];
-        w[c] = function(method: string, ...args: any[]): void { return clarity[method](...args); }
-        w[c].v = clarity.version;
-        while (queue.length > 0) { w[c](...queue.shift()); }
-    }
+    queue.clear();
```
toby-walker, line 5: >clear [](http://example.com/codeflow?start=10&length=5)

Minor; To me clear() would imply no processing. I think you process queued events. "process" or similar name would be closer to what this does.

packages/clarity-js/src/layout/mutation.ts
```diff
@@ -124,12 +124,13 @@ async function process(): Promise<void> {
   task.start(timer);
   while (mutations.length > 0) {
     let record = mutations.shift();
+    let instance = time();
```
toby-walker, line 127: >instance [](http://example.com/codeflow?start=8&length=8)

Minor: Name is odd-seeming to me as instance to me implies some sort of object, not a time. Did you mean instant?

packages/clarity-js/src/layout/mutation.ts
```diff
@@ -169,19 +170,22 @@ function track(m: MutationRecord, timer: Timer): string {
...
     // Initialize an entry if it doesn't already exist
-    history[key] = key in history ? history[key] : [0];
+    history[key] = key in history ? history[key] : [0, instance];
     let h = history[key];
     // Lookup any pending nodes queued up for removal, and process them now if we suspended a mutation before
-    if (inactive === false && h[0] >= Setting.MutationSuspendThreshold) { processNodeList(h[1], Source.ChildListRemove, timer); }
+    if (inactive === false && h[0] >= Setting.MutationSuspendThreshold) { processNodeList(h[2], Source.ChildListRemove, timer); }
     // Update the counter
-    h[0] = inactive ? h[0] + 1 : 1;
+    h[0] = inactive ? (h[1] === instance ? h[0] : h[0] + 1) : 1;
```
toby-walker, line 178: Style Suggestion: Add a brief comment to indicate what h[0/1/2] represent. Maybe declare constants for the indices to indicate what each represents.

packages/clarity-js/src/layout/mutation.ts
```diff
@@ -169,19 +170,22 @@ function track(m: MutationRecord, timer: Timer): string {
     // In those cases, IDs will change however the selector (which is relative to DOM xPath) remains the same
     let key = [parent, element, m.attributeName, names(m.addedNodes), names(m.removedNodes)].join();
     // Initialize an entry if it doesn't already exist
-    history[key] = key in history ? history[key] : [0];
+    history[key] = key in history ? history[key] : [0, instance];
```
toby-walker, line 173: Minor; Should  this really be three thing as h[2] stores list of nodes.

packages/clarity-js/src/layout/mutation.ts
toby-walker, line 178: Add comment about what this logic is handling.

toby-walker (approved):
:shipit:

packages/clarity-js/src/global.ts
sarveshnagpal: Good point. Done.

packages/clarity-js/src/layout/mutation.ts
sarveshnagpal: It's defined for the type "MutationHistory" with comments in layout.d.ts.

sarveshnagpal: The comment above is accurate -- that it is updating the counter.

packages/clarity-js/src/layout/mutation.ts
sarveshnagpal: As defined in the type... the last parameter is optional. So we don't have to explicitly give a null value. From the layout.d.ts -> export interface MutationHistory {
    [key: string]: [/* Count */ number, /* Instance */ number, /* Remove Nodes Buffer */ NodeList?];
}

packages/clarity-js/src/layout/mutation.ts
sarveshnagpal: It may not be the best choice for the word but I did mean instance -- as in single occurrence of something. The idea is that we don't want to suspend multiple mutations which occur together since that wouldn't really generate any additional chatter over the wire.

PR #347, opened 2023-02-15
Updating all packages and building insights bundle

sarveshnagpal:
- Updating all packages
- Updating rollup configuration to build an insights bundle
- Visualization bug fixes to better handle shadow DOM mutations 
- Bumping version to 0.7.3


packages/clarity-decode/package.json
```diff
@@ -26,24 +26,24 @@
...
+    "@rollup/plugin-node-resolve": "^15.0.0",
+    "@rollup/plugin-terser": "^0.4.0",
+    "@rollup/plugin-typescript": "^11.0.0",
+    "del-cli": "^5.0.0",
+    "husky": "^8.0.0",
+    "lint-staged": "^13.1.0",
+    "rollup": "^3.0.0",
     "ts-node": "^10.1.0",
     "tslint": "^6.1.3",
     "typescript": "^4.3.5"
```
toby-walker, line 42: This is pretty old. [https://github.com/microsoft/TypeScript/releases/tag/v4.3.5](https://github.com/microsoft/TypeScript/releases/tag/v4.3.5). From Jan 2021. 
Should we upgrade to current (4.9.5)?

packages/clarity-visualize/src/interaction.ts
```diff
@@ -1,5 +1,6 @@
 import { Asset, Constant, PlaybackState, Point, Setting } from "@clarity-types/visualize";
-import { Data, Interaction, Layout } from "clarity-decode";
+import { Data, Layout } from "clarity-js";
+import type { Interaction } from "clarity-decode"
```
toby-walker, line 3: Very minor: Use Decoded as prefix in all case to be consistent with the other changes you made.

packages/clarity-decode/package.json
sarveshnagpal: Good catch, I will address this in a subsequent change.

PR #364, opened 2023-03-19
fix(clarity-js): replaces hostname with host containing port

bertoluzza:
This pull request proposes some changes to improve the functionality of clarity-js on websites deployed on ports different than port 80 and port 443.

Currently, version 0.7.4 of clarity-js does not work properly on websites deployed on ports other than 80 and 443. For instance, when using clarity-js on a website deployed on port 8443, recordings and heatmaps won't work properly because clarity will try to download static assets from `example.com` instead of `example.com:8443`.

After reviewing the code, I found out that `hostname` was used in three instances instead of host, which causes the issues described above. Therefore, this pull request suggests replacing `hostname` with host in the relevant code blocks.

The proposed changes are backwards compatible because if no port is specified, `host` will have the same value as `hostname`.

### DIFF
--- a/packages/clarity-js/src/layout/node.ts
+++ b/packages/clarity-js/src/layout/node.ts
@@ -128,7 +128,7 @@ export default function (node: Node, source: Source): Node {
                 case "HEAD":
                     let head = { tag, attributes };
                     let l = insideFrame && node.ownerDocument?.location ? node.ownerDocument.location : location;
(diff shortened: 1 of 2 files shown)


bertoluzza:

sarveshnagpal (approved):
Thank you for fixing it @bertoluzza!

PR #372, opened 2023-04-04
Samart/api fixes

ender336:
Adding extract handling for hashes, also small bug fixes for metadata callbacks executing when we've already upgraded the session from lean and start never getting called once stop is called externally

### DIFF
--- a/packages/clarity-js/src/data/extract.ts
+++ b/packages/clarity-js/src/data/extract.ts
@@ -65,8 +65,8 @@ export function clone(v: Syntax[]): Syntax[] {
 
 export function compute(): void {
     try {
-        for (let v in variables) {
-            let key = parseInt(v);
+        for (let h in hashes) {
+            let key = parseInt(h);
             if (!(key in keys)) {
                 let variableData = variables[key];
                 for (let v in variableData) {
@@ -95,8 +95,8 @@ export function compute(): void {
                         for (let extractData of data[key]) {
                             if (extractData[0] === hashKey) {
                                 // extractData[1] is the value we uploaded last, check if it has changed
-                                if (extractData[1] !== content) {
-                                    shouldUpdate == true;
+                                if (extractData[1] === content) {
+                                    shouldUpdate = false;
                                 }
                             }
                         }
--- a/packages/clarity-js/src/data/extract.ts
+++ b/packages/clarity-js/src/data/extract.ts
@@ -3,9 +3,12 @@ import { Event, Setting, ExtractData } from "@clarity-types/data";
 import encode from "./encode";
 import * as internal from "@src/diagnostic/internal";
 import { Code, Constant, Severity } from "@clarity-types/data";
+import { hashText } from "@src/clarity";
 
 export let data: ExtractData = {};
 export let keys: number[] = [];
+// hash data is remembered so we can reupload once it changes
+let hashDataCache: { [key: string]: string } = {};
 
 let variables : { [key: number]: { [key: number]: Syntax[] }} = {};
 let selectors : { [key: number]: { [key: number]: string }} = {};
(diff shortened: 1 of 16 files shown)


packages/clarity-js/src/data/extract.ts
```diff
@@ -67,6 +88,19 @@ export function compute(): void {
...
+        for (let h in hashes) {
+            let key = parseInt(h);
+            let hashData = hashes[key];
+            for (let d in hashData) {
+                let hashKey = parseInt(d);
+                let content = hashText(hashData[hashKey]);
+                // only update when the contents have changes since we last checked
+                if (content != hashDataCache[hashData[hashKey]]) {
+                    hashDataCache[hashData[hashKey]] = content;
+                    update(key, hashKey, content.substring(0, Setting.ExtractLimit));
```
swaathee, line 100: instead of maintaining a seperate dictionary - pass in a additional parameter like force which if true would update if value has changed in data dictionary

packages/clarity-js/src/data/metadata.ts
```diff
@@ -85,7 +85,8 @@ export function stop(): void {
 }
 
 export function metadata(cb: MetadataCallback, wait: boolean = true): void {
-  if (data && wait === false) {
+  let upgraded = config.lean ? BooleanFlag.False : BooleanFlag.True;
+  if (data && (upgraded || wait === false)) {
```
swaathee, line 89: this is already done in processCallBack function why do we need this here?

packages/clarity-js/src/data/extract.ts
```diff
@@ -46,46 +65,64 @@ export function clone(v: Syntax[]): Syntax[] {
...
+
+            let hashData = hashes[key];
+            for (let h in hashData) {
+                let hashKey = parseInt(h);
+                let content = hashText(hashData[hashKey]).trim().substring(0, Setting.ExtractLimit);
+                newDataToUpload = newDataToUpload || update(key, hashKey, content);
+            }            
+        }
+
+        if (newDataToUpload) {
```
swaathee, line 96: can you call encode when keys is not empty instead?

packages/clarity-js/src/queue.ts
```diff
@@ -9,6 +9,7 @@ export function setup() {
     if (typeof w !== "undefined") {
         w[c] = function() {
             (w[c].q = w[c].q || []).push(arguments);
+            arguments[0] === "start" && w[c].q.unshift(w[c].q.pop()) && process();
```
toby-walker, line 12: What is this doing? Add comment explaining.

toby-walker:
>   callbacks.push({callback: cb, wait: wait });

If you call it immediately why do you add it to the list of callbacks? Is the idea that the callback would be called multiple times? For example if wait==false but lean is true then upgrade is called then there would be two callback invocations with different data parameter (False first ,true second)?

---
Refers to: packages/clarity-js/src/data/metadata.ts:96 in f06f9a6. [](commit_id = f06f9a6f91d818c25722ce864d46c64c70dc06e5, deletion_comment = False)

ender336:
> > callbacks.push({callback: cb, wait: wait });
> 
> If you call it immediately why do you add it to the list of callbacks? Is the idea that the callback would be called multiple times? For example if wait==false but lean is true then upgrade is called then there would be two callback invocations with different data parameter (False first ,true second)?
> 
> Refers to: packages/clarity-js/src/data/metadata.ts:96 in [f06f9a6](https://github.com/microsoft/clarity/commit/f06f9a6f91d818c25722ce864d46c64c70dc06e5). [](commit_id = [f06f9a6](https://github.com/microsoft/clarity/commit/f06f9a6f91d818c25722ce864d46c64c70dc06e5), deletion_comment = False)

Good catch, we shouldn't add it

packages/clarity-js/types/data.d.ts
```diff
@@ -409,7 +410,7 @@ export interface UpgradeData {
 }
 
 export interface ExtractData {
-    [key: number]: [number, string][]; // Array of [id, value] for every extracted data
+    [key: number]: { [subkey : number]: string }; // Array of [id, value] for every extracted data
```
swaathee, line 413: can you change the comment

swaathee (approved):

toby-walker (approved):
:shipit:

PR #380, opened 2023-04-30
Add DOB to clarity cookie

alaa-megahed:
- Update Clarity cookie to include DOB, to support new vs. returning users feature. 
- Add DOB to the envelope.
- Upgrade packages versions.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -2,7 +2,7 @@
   "packages": [
     "packages/*"
   ],
-  "version": "0.7.7",
+  "version": "0.7.8",
   "npmClient": "yarn",
   "useWorkspaces": true
 }
--- a/lerna.json
+++ b/lerna.json
@@ -2,7 +2,7 @@
   "packages": [
     "packages/*"
   ],
-  "version": "0.7.6",
+  "version": "0.7.7",
   "npmClient": "yarn",
   "useWorkspaces": true
 }
--- a/packages/clarity-decode/src/data.ts
+++ b/packages/clarity-decode/src/data.ts
@@ -80,7 +80,7 @@ export function decode(tokens: Data.Token[]): DataEvent {
                     let values = tokens[e++] as (number | string)[];
                     extract[key] = [];
                     for (let i = 0; i < values.length - 1; i += 2) {
-                        extract[key].push([values[i] as number, values[i + 1] as string]);
+                        extract[key][values[i] as number] = values[i + 1] as string;
                     }
                 }
             }
--- a/packages/clarity-decode/src/data.ts
+++ b/packages/clarity-decode/src/data.ts
@@ -101,6 +101,7 @@ export function envelope(tokens: Data.Token[]): Data.Envelope {
         sessionId: tokens[6] as string,
         pageNum: tokens[7] as number,
         upload: tokens[8] as Data.Upload,
-        end: tokens[9] as Data.BooleanFlag
+        end: tokens[9] as Data.BooleanFlag,
+        dob: tokens[10] as string
     };
 }
--- a/packages/clarity-decode/src/data.ts
+++ b/packages/clarity-decode/src/data.ts
@@ -101,7 +101,6 @@ export function envelope(tokens: Data.Token[]): Data.Envelope {
         sessionId: tokens[6] as string,
         pageNum: tokens[7] as number,
         upload: tokens[8] as Data.Upload,
-        end: tokens[9] as Data.BooleanFlag,
-        dob: tokens[10] as string
+        end: tokens[9] as Data.BooleanFlag
     };
 }
--- a/packages/clarity-js/src/data/metadata.ts
+++ b/packages/clarity-js/src/data/metadata.ts
@@ -85,11 +85,16 @@ export function stop(): void {
 }
 
 export function metadata(cb: MetadataCallback, wait: boolean = true): void {
-  if (data && wait === false) {
+  let upgraded = config.lean ? BooleanFlag.False : BooleanFlag.True;
+  // if caller hasn't specified that they want to skip waiting for upgrade but we've already upgraded, we need to
+  // directly execute the callback rather than adding to our list as we only process callbacks at the moment
+  // we go through the upgrading flow.
+  if (data && (upgraded || wait === false)) {
     // Immediately invoke the callback if the caller explicitly doesn't want to wait for the upgrade confirmation
     cb(data, !config.lean);
+  } else {
+    callbacks.push({callback: cb, wait: wait });
   }
-  callbacks.push({callback: cb, wait: wait });
 }
 
 export function id(): string {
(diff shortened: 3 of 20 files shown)


packages/clarity-decode/src/data.ts
```diff
@@ -101,6 +101,7 @@ export function envelope(tokens: Data.Token[]): Data.Envelope {
         sessionId: tokens[6] as string,
         pageNum: tokens[7] as number,
         upload: tokens[8] as Data.Upload,
-        end: tokens[9] as Data.BooleanFlag
+        end: tokens[9] as Data.BooleanFlag,
+        dob: tokens[10] as string
```
toby-walker, line 105: If this is in the cookie why needed here?

alaa-megahed: It's a 1P cookie, not being sent to the server.

packages/clarity-js/src/data/metadata.ts
```diff
@@ -144,9 +144,13 @@ function track(u: User, consent: BooleanFlag = null): void {
   // Convert time precision into days to reduce number of bytes we have to write in a cookie
   // E.g. Math.ceil(1628735962643 / (24*60*60*1000)) => 18852 (days) => ejo in base36 (13 bytes => 3 bytes)
   let end = Math.ceil((Date.now() + (Setting.Expire * Time.Day))/Time.Day);
+  // If dob is not set in the user object, use today's date as a dob
+  let date = new Date();
+  let dob = [date.getFullYear(), ('0' + (date.getMonth() + 1)).slice(-2), ('0' + date.getDate()).slice(-2)].join('');
```
toby-walker, line 149: Wouldn't this be in local time? Want to ensure in UTC. Instead use:

new Date().toISOString().Substring(0, 10).Replace('-','')

or similar. Definitely fix this to use UTC as that is critical.

packages/clarity-js/src/data/metadata.ts
```diff
@@ -144,9 +144,13 @@ function track(u: User, consent: BooleanFlag = null): void {
...
+  dob = u.dob === null ? dob : u.dob;
   // To avoid cookie churn, write user id cookie only once every day
-  if (u.expiry === null || Math.abs(end - u.expiry) >= Setting.CookieInterval || u.consent !== consent) {
-    setCookie(Constant.CookieKey, [data.userId, Setting.CookieVersion, end.toString(36), consent].join(Constant.Pipe), Setting.Expire);
+  if (u.expiry === null || Math.abs(end - u.expiry) >= Setting.CookieInterval || u.consent !== consent || u.dob !== dob) {
+    setCookie(Constant.CookieKey, [data.userId, Setting.CookieVersion, end.toString(36), consent, dob].join(Constant.Pipe), Setting.Expire);
```
toby-walker, line 153: Check that new cookie will work with older code. Can't assume that older client doesn't get new cookie.

packages/clarity-js/src/data/metadata.ts
toby-walker, line 149: Nit: Don't compute this unless it's needed.

toby-walker (changes requested):
:clock1:

packages/clarity-js/src/data/metadata.ts
alaa-megahed: Nice catch, updated the code.

packages/clarity-js/src/data/metadata.ts
alaa-megahed: Yeah I tested that, the older client code will read the first 4 parts of the cookie and ignore the dob. Similarly with the server code, will just ignore the DOB field of the envelope.

lerna.json
```diff
@@ -2,7 +2,7 @@
   "packages": [
     "packages/*"
   ],
-  "version": "0.7.6",
+  "version": "0.7.7",
```
ender336, line 5: package updated to 0.7.7 this morning, make sure to go to .8

ender336 (changes requested):

ender336 (approved):

toby-walker (approved):
:shipit:

packages/clarity-js/src/data/metadata.ts
```diff
@@ -149,9 +150,11 @@ function track(u: User, consent: BooleanFlag = null): void {
   // Convert time precision into days to reduce number of bytes we have to write in a cookie
   // E.g. Math.ceil(1628735962643 / (24*60*60*1000)) => 18852 (days) => ejo in base36 (13 bytes => 3 bytes)
   let end = Math.ceil((Date.now() + (Setting.Expire * Time.Day))/Time.Day);
+  // If DOB is not set in the user object, use today's date as a DOB
+  let dob = u.dob === null ? new Date().toISOString().substring(0, 10).replace(/-/g,'') : u.dob;
```
sarveshnagpal, line 154: Couple of points here: 
- This will convert the date to UTC and might not align with server date: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString
- Since these bytes will be added to every cookie we should consider compression. The easiest it to turn into base36 but we can also subtract it from a pivot date and then base36 it (e.g. current date - 20200101) and only send the delta.

sarveshnagpal (changes requested):

packages/clarity-js/src/data/metadata.ts
alaa-megahed: I have created a PR for ingest, could you take a look and review this one again please.

packages/clarity-js/src/data/metadata.ts
```diff
@@ -208,6 +213,7 @@ function user(): User {
     if (parts.length > 2) { output.expiry = num(parts[2], 36); }
     // Check if we have explicit consent to track this user
     if (parts.length > 3 && num(parts[3]) === 1) { output.consent = BooleanFlag.True; }
+    if (parts.length > 4 && num(parts[1]) > 1) { output.dob = parts[4]; }
```
swaathee, line 216: can you also log the version in dimension?

packages/clarity-js/src/data/metadata.ts
```diff
@@ -36,10 +36,11 @@ export function start(): void {
   dimension.log(Dimension.PageLanguage, document.documentElement.lang);
   dimension.log(Dimension.DocumentDirection, document.dir);
   dimension.log(Dimension.DevicePixelRatio, `${window.devicePixelRatio}`);
-  
+  dimension.log(Dimension.Dob, u.dob);
```
swaathee, line 39: does this need a null check

sarveshnagpal (approved):

packages/clarity-js/src/data/metadata.ts
alaa-megahed: The function already checks for null values first.

packages/clarity-js/src/data/metadata.ts
alaa-megahed: But why would the server need the version of the cookie?

PR #387, opened 2023-05-09
Improve identify API interface

sarveshnagpal:
- Automatically SHA-256 hash custom userId when used with identify API
- Bug fix to handle Electron apps


packages/clarity-js/src/data/variable.ts
```diff
@@ -13,10 +14,28 @@ export function set(variable: string, value: string | string[]): void {
     log(variable, values);
 }
 
-export function identify(userId: string, sessionId: string = null, pageId: string = null): void {
-    log(Constant.UserId, [userId]);
-    log(Constant.SessionId, [sessionId]);
-    log(Constant.PageId, [pageId]);
+export async function identify(userId: string, sessionId: string = null, pageId: string = null, userHint: string = null): Promise<IdentityData> {
+    let output: IdentityData = { userId: await sha256(userId), userHint: userHint || redact(userId) };
```
swaathee, line 18: null check for userId

packages/clarity-js/src/data/variable.ts
```diff
@@ -44,3 +63,22 @@ export function reset(): void {
...
+function redact(input: string): string {
+    return input.length >= Setting.WordLength ? 
+        `${input.substring(0,2)}${scrub(input.substring(2), Constant.Asterix, Constant.Asterix)}` : scrub(input, Constant.Asterix, Constant.Asterix);
+}
+
+async function sha256(input: string): Promise<string> {
+    try {
+        if (crypto) {
+            // Reference: https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest#converting_a_digest_to_a_hex_string
+            const buffer = await crypto.subtle.digest(Constant.SHA256, new (TextEncoder as any)(Constant.UTF8).encode(input));
```
keyrun, line 76: The default for TextEncoder is UTF8 and I think they removed support for everything else. So, you can call it without passing UTF8 and avoid the type casting to any.

keyrun (approved):

packages/clarity-js/src/data/variable.ts
sarveshnagpal: Done

packages/clarity-js/src/data/variable.ts
sarveshnagpal: Good point, fixed.

PR #404, opened 2023-06-13
Fix stylesheet and font loading

sosalah:
- Properly await on all stylesheets and fonts in the lists.
- Bump version to 0.7.9

### DIFF
--- a/packages/clarity-devtools/package.json
+++ b/packages/clarity-devtools/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clarity-devtools",
-  "version": "0.7.8",
+  "version": "0.7.9",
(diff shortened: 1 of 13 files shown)


packages/clarity-devtools/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clarity-devtools",
-  "version": "0.7.8",
```
swaathee, line 3: there are 2 more places where we have the version - version.js and the manifest.json can you please change all this?

sosalah: Thanks Swaathee, done

swaathee (approved):

PR #432, opened 2023-07-31
support for an image proxy service

ender336:


lerna.json
```diff
@@ -2,7 +2,7 @@
   "packages": [
     "packages/*"
   ],
-  "version": "0.7.8",
+  "version": "0.7.9",
```
swaathee, line 5: we are in version 0.7.9 already. sync issue?

swaathee (changes requested):

swaathee (approved):

PR #446, opened 2023-08-21
V alihashish/actions

AliHashish:
Executing a few customizable actions depending on some signals:
- 2 default actions are provided, one with a light and dark mode theme (easily extensible to accomodate more default actions).
- Custom actions can be inserted by the client (only concern is use of eval() function to execute client's code - possible security concern?)
- Accepts extra parameters


AliHashish:

PR #451, opened 2023-08-30
Samart/animate support

ender336:


Ambekhie (approved):

packages/clarity-js/src/layout/animation.ts
```diff
@@ -0,0 +1,77 @@
...
+    state = [];
+    reset();
+}
+
+function overrideAnimationHelper(whereToStoreFunction: () => void, name: string) {
+    if (whereToStoreFunction === null) {
+      whereToStoreFunction = Animation.prototype[name];
+      Animation['clarity'] = true;
+      Animation.prototype[name] = function(): void {
+        if (!this['clarityAnimationName']) {
```
keyrun, line 53: >clarityAnimationName [](http://example.com/codeflow?start=19&length=20)

Would clarityAnimationId be better? Also think it would best to store this as a const outside since the same string is used in many places.

packages/clarity-js/src/layout/animation.ts
```diff
@@ -0,0 +1,77 @@
...
+    overrideAnimationHelper(animationPause, "pause");
+    overrideAnimationHelper(animationCancel, "cancel");
+    overrideAnimationHelper(animationFinish, "finish");
+}
+
+export function reset(): void {
+    state = [];
+}
+
+export function track(time: number, id: string, operation: AnimationOperation, keyFrames?: string, timing?: string, targetId?: number, timeline?: string): void {
```
keyrun, line 26: >export [](http://example.com/codeflow?start=0&length=6)

Does this have to be exported?

packages/clarity-js/src/layout/animation.ts
```diff
@@ -0,0 +1,77 @@
...
+    encode(Event.Animation);
+}
+
+export function stop(): void {
+    state = [];
+    reset();
+}
+
+function overrideAnimationHelper(whereToStoreFunction: () => void, name: string) {
+    if (whereToStoreFunction === null) {
```
swaathee, line 49: you need to add core.active() check to make sure that we are not executing any clarity specific logic if clarity is paused

packages/clarity-js/src/layout/animation.ts
```diff
@@ -0,0 +1,77 @@
...
+                track(time(), this['clarityAnimationName'], AnimationOperation.Pause);
+                break;
+            case "cancel":
+                track(time(), this['clarityAnimationName'], AnimationOperation.Cancel);
+                break;
+            case "finish":
+                track(time(), this['clarityAnimationName'], AnimationOperation.Finish);
+                break;
+        }
+        return whereToStoreFunction.apply(this, arguments);
```
swaathee, line 74: rename to functionToOverride?

packages/clarity-js/src/layout/animation.ts
```diff
@@ -0,0 +1,77 @@
...
+function overrideAnimationHelper(whereToStoreFunction: () => void, name: string) {
+    if (whereToStoreFunction === null) {
+      whereToStoreFunction = Animation.prototype[name];
+      Animation['clarity'] = true;
+      Animation.prototype[name] = function(): void {
+        if (!this['clarityAnimationName']) {
+          this['clarityAnimationName'] = shortid();
+          let keyframes = (<KeyframeEffect>this.effect).getKeyframes();
+          let timing = (<KeyframeEffect>this.effect).getTiming();
+          track(time(), this['clarityAnimationName'], AnimationOperation.Create, JSON.stringify(keyframes), JSON.stringify(timing), getId(this.effect.target));
```
swaathee, line 57: should this logic be within play below? 
rather what happens if we get a pause event before play? how would we know what to pause.

packages/clarity-visualize/src/visualizer.ts
```diff
@@ -98,8 +98,8 @@ export class Visualizer implements VisualizerType {
                     for (let entry of p) {
                         switch (key) {
                             case Constant.Dom:
-                                let dom = this.enrich.selectors(entry);
                                 if (entry.event === Data.Event.Discover) {
+                                    let dom = this.enrich.selectors(entry);
```
swaathee, line 102: is this an intended change?

packages/clarity-js/src/layout/animation.ts
```diff
@@ -0,0 +1,77 @@
...
+
+export function stop(): void {
+    state = [];
+    reset();
+}
+
+function overrideAnimationHelper(whereToStoreFunction: () => void, name: string) {
+    if (whereToStoreFunction === null) {
+      whereToStoreFunction = Animation.prototype[name];
+      Animation['clarity'] = true;
```
swaathee, line 51: where is this used?

packages/clarity-js/src/layout/animation.ts
```diff
@@ -0,0 +1,77 @@
...
+      Animation.prototype[name] = function(): void {
+        if (!this['clarityAnimationName']) {
+          this['clarityAnimationName'] = shortid();
+          let keyframes = (<KeyframeEffect>this.effect).getKeyframes();
+          let timing = (<KeyframeEffect>this.effect).getTiming();
+          track(time(), this['clarityAnimationName'], AnimationOperation.Create, JSON.stringify(keyframes), JSON.stringify(timing), getId(this.effect.target));
+        }
+
+        switch (name) {
+            case "play":
```
swaathee, line 61: can you make sure that if someone pauses and plays infinite times on a loop - we stop collecting data

swaathee (changes requested):

packages/clarity-js/src/layout/animation.ts
ender336: the goal here is that if we haven't seen an animation before (clarityAnimationId isn't present) - we create the animation first.

Once it is created we execute the operation. It could be play or pause, but calling pause on a newly created animation is just a noop as far as the browser is concerned.

packages/clarity-visualize/src/visualizer.ts
ender336: I thought we didn't need to create the object when we didn't use it - but it does look like enrich has some side effects so I'll revert

packages/clarity-decode/src/layout.ts
```diff
@@ -1,6 +1,7 @@
 import { Constant } from "@clarity-types/data";
 import { Data, Layout } from "clarity-js";
 import { DomData, LayoutEvent, Interaction, RegionVisibility  } from "../types/layout";
+import { AnimationData } from "clarity-js/types/layout";
```
ender336, line 4: >AnimationData [](http://example.com/codeflow?start=9&length=13)

should make this Layout.AnimationData instead

packages/clarity-js/src/layout/animation.ts
```diff
@@ -0,0 +1,85 @@
...
+                this[operationCount] = 0;
+                let keyframes = (<KeyframeEffect>this.effect).getKeyframes();
+                let timing = (<KeyframeEffect>this.effect).getTiming();
+                track(time(), this[animationId], AnimationOperation.Create, JSON.stringify(keyframes), JSON.stringify(timing), getId(this.effect.target));
+            }
+
+            if (this[operationCount]++ < maxOperations)  {
+                switch (name) {
+                    case "play":
+                        track(time(), this[animationId], AnimationOperation.Play);
```
swaathee, line 67: nit -   assign a variable and have a single call to track to keep it clean.

swaathee (approved):

PR #465, opened 2023-09-25
Samart/css replace

ender336:
Handling adopted styles


packages/clarity-js/package.json
```diff
@@ -1,12 +1,13 @@
...
   "name": "clarity-js",
-  "version": "0.7.18",
+  "version": "0.7.19",
   "description": "An analytics library that uses web page interactions to generate aggregated insights",
   "author": "Microsoft Corp.",
   "license": "MIT",
   "main": "build/clarity.js",
   "module": "build/clarity.module.js",
   "unpkg": "build/clarity.min.js",
+  "special": "build/clarity.special.js",
```
sarveshnagpal, line 10: >special [](http://example.com/codeflow?start=28&length=7)

Calling it something other than special? extended?

packages/clarity-js/src/data/upload.ts
```diff
@@ -16,6 +16,7 @@ import * as ping from "@src/data/ping";
 import * as timeline from "@src/interaction/timeline";
 import * as region from "@src/layout/region";
 import * as extract from "@src/data/extract";
+import * as adoptedStyles from "@src/layout/adoptedStyles";
```
sarveshnagpal, line 19: >adoptedStyles [](http://example.com/codeflow?start=44&length=13)

Let's call it "styles" in line with rest of the naming

packages/clarity-js/src/layout/encode.ts
```diff
@@ -36,6 +37,25 @@ export default async function (type: Event, timer: Timer = null, ts: number = nu
             }
             region.reset();
             break;
+        case Event.StyleSheetAdoption:
```
sarveshnagpal, line 40: >Event.StyleSheetAdoption: [](http://example.com/codeflow?start=13&length=25)

Technically these are extra bytes that will make it to default bundle. I don't have a clean solution here but pointing out so we can think through this async (non-blocking comment).

sarveshnagpal (approved):
:shipit:

PR #483, opened 2023-10-11
Remove node references when they are removed

MustafaLeithy:


packages/clarity-js/src/layout/dom.ts
```diff
@@ -339,6 +340,12 @@ function remove(id: number, source: Source): void {
         value.metadata.active = false;
         value.parent = null;
         track(id, source);
+
+        // Clean up node references for removed nodes
+        nodesMap.delete(id);
+        for (let childId of value.children){
+            nodesMap.delete(childId);
+        }
```
ender336, line 348: does this need to be recursive to delete decedents below immediate children?

packages/clarity-js/src/layout/dom.ts
```diff
@@ -297,10 +297,11 @@ export function hashText(hash: string): string {
 }
 
 export function getNode(id: number): Node {
```
ender336, line 299: >getNode [](http://example.com/codeflow?start=16&length=7)

not sure it's any more efficient but feels a little more compact to have this be:

return nodesMap.has(id) ? nodesMap.get(id) : null

ender336 (approved):

PR #496, opened 2023-11-02
grabbing css content for electron apps rather than references

ender336:


packages/clarity-js/src/layout/node.ts
```diff
@@ -157,6 +158,21 @@ export default function (node: Node, source: Source): Node {
                     }
                     dom[call](node, parent, frameData, source);
                     break;
+                case "LINK":
+                    // electron stylesheets reference the local file system - translating those
+                    // to inline styles so playback can work
+                    if (electron && attributes['rel'] === 'stylesheet') {
```
sarveshnagpal, line 164: >stylesheet [](http://example.com/codeflow?start=59&length=10)

Worth moving this to Constant in d.ts file and then use it here like some of the other constants in this file.

packages/clarity-js/src/layout/node.ts
```diff
@@ -157,6 +158,21 @@ export default function (node: Node, source: Source): Node {
...
+                            var currentStyleSheet = document.styleSheets[styleSheetIndex];
+                            if (currentStyleSheet.ownerNode == element) {
+                                let syntheticStyleData = { tag: "STYLE", attributes, value: getCssRules(currentStyleSheet) };
+                                dom[call](node, parent, syntheticStyleData, source);
+                                break;
+                            }
+                        }
+                        break;
+                    }
+                    // intentionally not breaking here so we can fall through to default behavior when not dealing with electron
```
sarveshnagpal, line 175: >// intentionally not breaking h [](http://example.com/codeflow?start=20&length=31)

I think it's probably better to include two lines:
let linkData = 
dom[call](); 

and then call break so that in future if someone injects anything between LINK and default; they don't accidentally create bugs.

sarveshnagpal (approved):
:shipit:

PR #500, opened 2023-11-07
Uri encode cookie values for compatibility with certain mobile browsers

MustafaLeithy:
Some browsers automatically url encode cookie values if they are not url encoded. We fix this by always uri encoding cookies, and supporting the decoding of encoded cookies.

### DIFF
--- a/packages/clarity-js/src/data/metadata.ts
+++ b/packages/clarity-js/src/data/metadata.ts
@@ -282,12 +282,12 @@ function setCookie(key: string, value: string, time: number): void {
   if (config.track && ((navigator && navigator.cookieEnabled) || supported(document, Constant.Cookie))) {
     // Some browsers automatically url encode cookie values if they are not url encoded.
     // We therefore encode and decode cookie values ourselves.
-    value = encodeCookieValue(value);
+    let encodedValue = encodeCookieValue(value);
 
     let expiry = new Date();
     expiry.setDate(expiry.getDate() + time);
     let expires = expiry ? Constant.Expires + expiry.toUTCString() : Constant.Empty;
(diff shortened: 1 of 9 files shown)


packages/clarity-js/src/data/metadata.ts
```diff
@@ -243,16 +243,54 @@ function getCookie(key: string): string {
...
 
+function decodeCookieValue(value: string): string {
+  // For backwards compatability we need to consider 3 cases:
+  // * Cookie was previously not encoded by Clarity and browser did not encode it
+  // * Cookie was previously not encoded by Clarity and browser encoded it once or more
+  // * Cookie was previously encoded by Clarity and browser did not encode it
+  let [isEncoded, decodedValue] = valueIsEncoded(value);
+  if (!isEncoded) {
+    return value;
+  }
```
ender336, line 264: we can save some bytes by cutting lines 262-264, while loop will be skipped and decodedValue would be returned

packages/clarity-js/src/data/metadata.ts
```diff
@@ -243,16 +243,54 @@ function getCookie(key: string): string {
...
+
+  while (isEncoded) {
+    [isEncoded, decodedValue] = valueIsEncoded(decodedValue);
+  }
+
+  return decodedValue;
+}
+
+function valueIsEncoded(value: string): [boolean, string] {
```
ender336, line 273: >valueIsEncoded [](http://example.com/codeflow?start=9&length=14)

nit: this name feels a little deceptive as it actually does the work to decode the cookie. Do you think it would be more clear if we inlined what you currently have as decodeCookieValue into the getCookieFunction (it's pretty small) and renamed valueIsEncoded to be decodeCookieValue?

ender336 (approved):
:shipit:

PR #501, opened 2023-11-07
Enable the customization of the name of DOM elements when clicked

jonathanghm:
**Objective**
Enable the customization of the name of DOM elements when clicked.

**Motivation**
**1. Independence from the content of the DOM element:** 
By using a custom attribute, the logic for event tracking is separated from the content of the element. This provides flexibility to change the content without affecting the tracking logic.

**2. Avoids internationalization issues:** 
In multilingual applications, the text of the element may change depending on the language, which could complicate tracking. A custom attribute can provide a more stable identifier.

**2. Privacy enhancement:**
There may be scenarios where an element contains sensitive information, like email, name or passwords. This enhances the masking option already provided by clarity and additionally gives the analytics more context about the action.

**How does it work?**
By using the custom attribute "**data-clarity-event**" on the clickable DOM element, this will be considered as the default event name.
If the custom attribute "**data-clarity-event**" does not exist on the element, the current logic will be followed to obtain the name (textContent, value, or alt).

### DIFF
--- a/packages/clarity-js/src/interaction/click.ts
+++ b/packages/clarity-js/src/interaction/click.ts
@@ -1,6 +1,6 @@
 import { BooleanFlag, Constant, Event, Setting } from "@clarity-types/data";
 import { BrowsingContext, ClickState } from "@clarity-types/interaction";
-import { Box } from "@clarity-types/layout";
+import { Box, Constant as LayoutConstant } from "@clarity-types/layout";
 import { bind } from "@src/core/event";
 import { schedule } from "@src/core/task";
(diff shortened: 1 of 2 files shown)


jonathanghm:

YoRab (approved):

DorianGentine (approved):

legz (approved):

legz:
This evolution would be very useful in our context (we have buttons without text labels that Clarity can understand).
Would it be possible to review this PR please?

PR #539, opened 2024-02-26
Adding noscript placeholders

ender336:
Some css relies on specific number of children or position of children within a DOM tree. Adding noscript tags into our visualized DOM to maintain correctness.

### DIFF
--- a/packages/clarity-js/src/layout/node.ts
+++ b/packages/clarity-js/src/layout/node.ts
@@ -108,6 +108,11 @@ export default function (node: Node, source: Source): Node {
                     }
                     break;
                 case "NOSCRIPT":
+                    // keeping the noscript tag but removing all of its contents. Some HTML markup relies on having these tags
+                    // to maintain parity with the original css view, but we don't want to execute any noscript in Clarity
+                    element.innerHTML = '';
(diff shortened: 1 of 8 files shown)


packages/clarity-js/src/layout/node.ts
```diff
@@ -108,6 +108,11 @@ export default function (node: Node, source: Source): Node {
                     }
                     break;
                 case "NOSCRIPT":
+                    // keeping the noscript tag but removing all of its contents. Some HTML markup relies on having these tags
+                    // to maintain parity with the original css view, but we don't want to execute any noscript in Clarity
```
toby-walker, line 112: >maintain [](http://example.com/codeflow?start=26&length=8)

Just asking for us to think through: Why not send the content to avoid even more bizarre cases of CSS that refer to noscript tag's content?

toby-walker (approved):
:shipit:

packages/clarity-js/src/layout/node.ts
ender336: That's not possible (to the best of my knowledge anyway). Sending the content is wasteful anyway as javascript will always be enabled in playbacks (otherwise they won't play :P )

PR #550, opened 2024-03-12
Delay layout module experiment

AbdelrhmanMagdy:
delaying the layout module after DOM finish loading. This is part of our experiments that target optimizing the initial page load time.

### DIFF
--- a/packages/clarity-js/src/clarity.ts
+++ b/packages/clarity-js/src/clarity.ts
@@ -13,7 +13,6 @@ import * as performance from "@src/performance";
 export { version };
 export { consent, event, identify, set, upgrade, metadata } from "@src/data";
 export { hashText } from "@src/layout";
-import { bind } from "@src/core/event";
 
 const modules: Module[] = [diagnostic, layout, interaction, performance];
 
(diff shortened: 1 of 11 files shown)


AbdelrhmanMagdy:

packages/clarity-js/src/clarity.ts
```diff
@@ -22,7 +23,16 @@ export function start(config: Config = null): void {
     core.config(config);
     core.start();
     data.start();
-    modules.forEach(x => measure(x.start)());
+    modules.forEach((x, idx) => {
+      if (idx === 1) {
```
swaathee, line 27: I would separate out modules because checking based on index is confusing and error prone. 
Also the modules need to stop in reverse order of initialization - Can you check what the impact is when layout module is stopped out of order.

AbdelrhmanMagdy: I'm now checking on the module reference instead of the index.
Regarding your second point, my understanding is that this situation might only affect the diagnostic module based on its order in the array, which could depend on others. However, after conducting several experiments, I ensured that stopping the layout module before its initialization doesn't have any adverse effects.

swaathee (approved):

Issue #407, opened 2023-06-16
Memory leak issue


ohryk-intellias:
After adding Clarity through the GTM we faced the issue with memory leak. Seems that there are a lot of strings allocated in the memory and for some reason, it can't be cleaned up. After some time of usage of our search page we can see the following:

<img width="1150" alt="Screenshot 2023-06-16 at 17 09 22" src="https://github.com/microsoft/clarity/assets/64080949/96eed985-77d9-47a2-8d67-05a695840095">
<img width="1261" alt="Screenshot 2023-06-16 at 16 52 11" src="https://github.com/microsoft/clarity/assets/64080949/e51f46c0-bbeb-4d00-ae74-f7c19b7969fa">

And the context in which the retained objects are stored:

<img width="566" alt="Screenshot 2023-06-16 at 17 40 36" src="https://github.com/microsoft/clarity/assets/64080949/10736296-7791-45c2-b251-8733d9e8ccab">

jsmircic:
Same issue. for some reason `/collect` calls stop and after that app memory usage just grows until Out Of Memory crash.. Can't really detect when that happens but it might be related to browser tab idling for longer time. 
For now we will remove clarity integration as this is affecting customers. Will monitor this issue as /clarity was helpful for us.

ClaritySupport:
@ohryk-intellias , We are Clarity support.Could you please share below details to investigate the issue further.

1. Link to page on your website having this issue.
2. Steps to reproduce the issue.
3. Can you please send a test account with some test data so that we can see the memory usage and debug it?

zheller:
We ran into this issue as well (in segment + GTM integrations)

ClaritySupport:
@zheller ,We are Clarity support.Could you please share below details to investigate the issue further.

Link to page on your website having this issue.
Steps to reproduce the issue.
Can you please send a test account with some test data so that we can see the memory usage and debug it?

R-Pin3da:
I hit this issue too.
In my case it's happening on a page with an audio tag with `src="data:audio/mpeg;base64,...`. This base64 is about 100MB.
Seems that /collect is sending the whole content of the tag (3 times):
![image](https://github.com/microsoft/clarity/assets/54686431/13bb615e-9075-4e43-87b1-5afff2956c81)
![image](https://github.com/microsoft/clarity/assets/54686431/ab00154f-1cc8-476d-b247-46e95f31ccdc)

After that, the /collect endpoint is not called anymore.

[...]

PR #551, opened 2024-03-13
Ignoring src in media tags

AbdelrhmanMagdy:
Ignoring big base64 encoded src for video and audio html tags to prevent large payload tokens and memory leaks.
Addressing [this issue](https://github.com/microsoft/clarity/issues/407#issuecomment-1992055905)

### DIFF
--- a/packages/clarity-js/src/layout/node.ts
+++ b/packages/clarity-js/src/layout/node.ts
@@ -186,7 +186,7 @@ export default function (node: Node, source: Source): Node {
                 case "AUDIO":
                 case "SOURCE":
                     // Ignoring any base64 src attribute for media elements to prevent big unused tokens to be sent and shock the network 
-                    if(Constant.Src in attributes && attributes[Constant.Src].startsWith("data:")){
+                    if (Constant.Src in attributes && attributes[Constant.Src].startsWith("data:")) {
                         delete attributes[Constant.Src]
                     }
                     let mediaTag = { tag, attributes };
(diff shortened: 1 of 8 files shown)


packages/clarity-js/src/layout/node.ts
```diff
@@ -182,6 +182,16 @@ export default function (node: Node, source: Source): Node {
                     let linkData = { tag, attributes };
                     dom[call](node, parent, linkData, source);
                     break;
+                case "VIDEO":
+                case "AUDIO":
+                case "SOURCE":
+                    // Ignoring any base64 src attribute for media elements to prevent big unused tokens to be sent and shock the network 
+                    if(Constant.Src in attributes && attributes[Constant.Src].startsWith("data:")){
```
ender336, line 189: nit: missing a space after if

ender336 (approved):

packages/clarity-js/src/layout/node.ts
```diff
@@ -182,6 +182,16 @@ export default function (node: Node, source: Source): Node {
                     let linkData = { tag, attributes };
                     dom[call](node, parent, linkData, source);
                     break;
+                case "VIDEO":
+                case "AUDIO":
+                case "SOURCE":
+                    // Ignoring any base64 src attribute for media elements to prevent big unused tokens to be sent and shock the network 
+                    if (Constant.Src in attributes && attributes[Constant.Src].startsWith("data:")) {
+                        delete attributes[Constant.Src]
```
swaathee, line 190: do mediaTag.attributes[Constant.Src] = empty string instead of deleting here

swaathee: also how is empty src handled on visualiaze?

AbdelrhmanMagdy: An empty "src" attribute results in displaying an HTML media element, such as audio/video, with a playback length of 0.00.

PR #559, opened 2024-03-21
ensuring we clear out cookie values when consent is revoked

ender336:
Writing zero as our time was making it a session cookie, switching to negative max value to ensure it is expiring and adding some extra validation to ensure we don't start writing cookies again for subsequent calls until consent is captured.


MustafaLeithy (approved):
LGTM

swaathee (approved):

PR #564, opened 2024-03-29
stopping clarity from running if globalPrivacyControl is true

ender336:


packages/clarity-js/src/core/index.ts
```diff
@@ -41,7 +41,8 @@ export function check(): boolean {
             document["createTreeWalker"] &&
             "now" in Date &&
             "now" in performance &&
-            typeof WeakMap !== "undefined";
+            typeof WeakMap !== "undefined" &&
+            navigator['globalPrivacyControl'] !== true
```
swaathee, line 45: can you use (navigator && "(navigator && "globalPrivacyControl" in navigator") before testing for true - that is similar to how we extract navigator values in other places

AbdelrhmanMagdy (approved):

swaathee (approved):

PR #581, opened 2024-04-16
send first scrollevent

zoeyang7633:
- sending 1st scroll event for attention map data

### DIFF
--- a/packages/clarity-decode/src/data.ts
+++ b/packages/clarity-decode/src/data.ts
@@ -57,7 +57,8 @@ export function decode(tokens: Data.Token[]): DataEvent {
                 scrollY: tokens[8] as number,
                 pointerX: tokens[9] as number,
                 pointerY: tokens[10] as number,
-                activityTime: tokens[11] as number
+                activityTime: tokens[11] as number,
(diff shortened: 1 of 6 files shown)


packages/clarity-js/src/interaction/scroll.ts
```diff
@@ -50,8 +50,9 @@ function recompute(event: UIEvent = null): void {
 
     let current: ScrollState = { time: time(event), event: Event.Scroll, data: {target: element, x, y, top, bottom} };
 
+    // Commenting out this part to send the first scroll event for attention map
```
ender336, line 53: lets not check in commented out code - if you dont need it just delete it please

packages/clarity-decode/src/data.ts
```diff
@@ -57,8 +57,7 @@ export function decode(tokens: Data.Token[]): DataEvent {
                 scrollY: tokens[8] as number,
                 pointerX: tokens[9] as number,
                 pointerY: tokens[10] as number,
-                activityTime: tokens[11] as number,
-                scrollTime: tokens.length > 12 ? tokens[12] as number: null
```
swaathee, line 61: this is problematic since it can cause exceptions when we add another type field in the same position in future

PR #587, opened 2024-04-24
tracking initial top & bottom elements

zoeyang7633:


packages/clarity-decode/src/interaction.ts
```diff
@@ -74,7 +74,7 @@ export function decode(tokens: Data.Token[]): InteractionEvent {
                 x: tokens[3] as number,
                 y: tokens[4] as number,
                 top: tokens.length > 5 ? tokens[5] as string : null,
-                bottom: tokens.length > 6 ? tokens[6] as string : null
+                bottom: tokens.length > 6 ? tokens[6] as string : null,
```
swaathee, line 77: remove

packages/clarity-js/src/interaction/encode.ts
```diff
@@ -117,13 +117,15 @@ export default async function (type: Event, ts: number = null): Promise<void> {
...
+                const top = metadata(entry.data.top as Node, entry.event);
+                const bottom = metadata(entry.data.bottom as Node, entry.event);
                 if (sTarget.id > 0) {
                     tokens = [entry.time, entry.event];
                     tokens.push(sTarget.id);
                     tokens.push(entry.data.x);
                     tokens.push(entry.data.y);
-                    tokens.push(entry.data.top);
-                    tokens.push(entry.data.bottom);
+                    tokens.push(top?.hash?.[1]);
```
swaathee, line 127: please encode all hashes similar to click hash - also change decode

packages/clarity-js/src/layout/discover.ts
```diff
@@ -20,11 +23,19 @@ export function start(): void {
 
 async function discover(): Promise<void> {
     let ts = time();
-    
     let timer: Timer = { id: id(), cost: Metric.LayoutCost };
     task.start(timer);
     await traverse(document, timer, Source.Discover, ts);
+    logInitial();
```
swaathee, line 29: if user scrolls before traverse this wont be the initial scroll positions

packages/clarity-js/src/interaction/scroll.ts
```diff
@@ -62,7 +55,19 @@ function recompute(event: UIEvent = null): void {
     timeout = setTimeout(process, Setting.LookAhead, Event.Scroll);
 }
 
-function getPositionHash(x: number, y: number): string {
```
swaathee, line 65: prefer your previous approach to returning node arrays - it is more confusing

swaathee (changes requested):

packages/clarity-js/src/interaction/scroll.ts
```diff
@@ -93,6 +100,15 @@ function similar(last: ScrollState, current: ScrollState): boolean {
     return (dx * dx + dy * dy < Setting.Distance * Setting.Distance) && (current.time - last.time < Setting.Interval);
 }
 
+export function compute(): void {
+    if (initTopNode || initBottomNode) {
+        const sTopTarget = metadata(initTopNode, null);
+        const sBottomTarget = metadata(initBottomNode, null);
+        dimension.log(Dimension.InitialTop, sTopTarget?.hash?.[1]);
```
swaathee, line 107: send all hashes

packages/clarity-decode/src/interaction.ts
```diff
@@ -69,12 +69,14 @@ export function decode(tokens: Data.Token[]): InteractionEvent {
...
         case Data.Event.Scroll:
+            const topHash = tokens.length > 5 ? (tokens[5] as string).split(Data.Constant.Dot) : null;
+            const bottomHash = tokens.length > 6 ? (tokens[6] as string).split(Data.Constant.Dot) : null;
             let scrollData: Interaction.ScrollData = {
                 target: tokens[2] as number,
                 x: tokens[3] as number,
                 y: tokens[4] as number,
-                top: tokens.length > 5 ? tokens[5] as string : null,
-                bottom: tokens.length > 6 ? tokens[6] as string : null
+                top: topHash?.length > 1 ? topHash[1] : null,
```
ender336, line 78: >topHash?.length > 1 ? topHash[1] : null [](http://example.com/codeflow?start=21&length=39)

are you doing this to get the alpha hash instead of beta (or vice versa) ? You should upload both and handle that server side otherwise you'll break as soon as anyone starts flighting a new hashing algorithm

packages/clarity-js/src/interaction/encode.ts
```diff
@@ -117,13 +117,17 @@ export default async function (type: Event, ts: number = null): Promise<void> {
         case Event.Scroll:
             for (let entry of scroll.state) {
                 let sTarget = metadata(entry.data.target as Node, entry.event);
+                const top = metadata(entry.data.top as Node, entry.event);
+                const bottom = metadata(entry.data.bottom as Node, entry.event);
+                const cTopHash = top.hash ? top.hash.join(Constant.Dot) : Constant.Empty;
```
ender336, line 122: >cTopHash [](http://example.com/codeflow?start=22&length=8)

sTopHash / sBottomHash right? The s stands for scroll and I think you grabbed these from the click section

ender336, line 122: >top [](http://example.com/codeflow?start=33&length=3)

if there is any risk of top or bottom being null you may want to do top?.hash and bottom?.

packages/clarity-js/src/interaction/scroll.ts
```diff
@@ -1,14 +1,17 @@
...
 import { clearTimeout, setTimeout } from "@src/core/timeout";
-import { iframe, get } from "@src/layout/dom";
-import { target } from "@src/layout/target";
+import { iframe } from "@src/layout/dom";
+import { target, metadata } from "@src/layout/target";
 import encode from "./encode";
+import * as dimension from "@src/data/dimension";
 
 export let state: ScrollState[] = [];
+let initTopNode: Node;
```
ender336, line 13: >initTopNode [](http://example.com/codeflow?start=4&length=11)

nit: I'd really prefer initialTopNode to initTopNode, it sounds like you are going to initialize something

packages/clarity-js/src/interaction/scroll.ts
```diff
@@ -45,13 +48,17 @@ function recompute(event: UIEvent = null): void {
...
+    const top = getPositionNode(xPosition, startYPosition);
+    const bottom = getPositionNode(xPosition, endYPosition);
 
     let current: ScrollState = { time: time(event), event: Event.Scroll, data: {target: element, x, y, top, bottom} };
 
     // We don't send any scroll events if this is the first event and the current position is top (0,0)
-    if ((event === null && x === 0 && y === 0) || (x === null || y === null)) { return; }
+    if ((event === null && x === 0 && y === 0) || (x === null || y === null)) {
+        initTopNode = top;
+        initBottomNode = bottom;
```
ender336, line 59: I thought part of the issue here was that the scroll event didn't fire until the scroll had actually completed. Don't you need to grab the initial top/bottom nodes before any scroll events?

packages/clarity-js/src/interaction/scroll.ts
ender336, line 107: >sTopTarget?.hash?.[1] [](http://example.com/codeflow?start=44&length=21)

also here, I don't think you should only look at one of the two hashes

packages/clarity-js/src/interaction/scroll.ts
```diff
@@ -93,6 +100,15 @@ function similar(last: ScrollState, current: ScrollState): boolean {
     return (dx * dx + dy * dy < Setting.Distance * Setting.Distance) && (current.time - last.time < Setting.Interval);
 }
 
+export function compute(): void {
+    if (initTopNode || initBottomNode) {
```
ender336, line 104: >initTopNode || initBottomNode [](http://example.com/codeflow?start=8&length=29)

nit but if you are going to have this outer check you should split it into two separate if clauses (one for top and one for bottom)

packages/clarity-js/src/interaction/scroll.ts
```diff
@@ -93,6 +100,15 @@ function similar(last: ScrollState, current: ScrollState): boolean {
...
+        const sTopTarget = metadata(initTopNode, null);
```
ender336, line 105: >sTopTarget [](http://example.com/codeflow?start=14&length=10)

nit: drop the s prefix here

ender336 (changes requested):
:clock1:

packages/clarity-js/types/data.d.ts
```diff
@@ -145,7 +145,9 @@ export const enum Dimension {
     ConnectionType = 27,
     Dob = 28,
     CookieVersion = 29,
-    DeviceFamily = 30 // Allows iOS SDK to override the DeviceFamily value parsed from UserAgent.
+    DeviceFamily = 30, // Allows iOS SDK to override the DeviceFamily value parsed from UserAgent.
+    InitialTop = 31,
```
swaathee, line 149: nit: InitialScrollTop or just ScrollTop might be clearer

packages/clarity-js/src/interaction/scroll.ts
swaathee: you can also drop node and just do initialTop and Bottom

packages/clarity-js/src/interaction/scroll.ts
```diff
@@ -76,7 +83,7 @@ function getPositionHash(x: number, y: number): string {
         node = node.parentNode;
     }
 
-    return get(node)?.hash?.[1];
+    return node;
 }
 
 export function reset(): void {
```
swaathee, line 89: you should reset top and bottom initial nodes both here and in stop

packages/clarity-js/src/interaction/encode.ts
zoeyang7633: metadata returns a object by default so top or bottom won't be null!

packages/clarity-js/src/interaction/scroll.ts
zoeyang7633: this gets called when the scroll module starts, before any scroll event happens

packages/clarity-js/src/layout/discover.ts
```diff
@@ -10,17 +10,18 @@ import encode from "@src/layout/encode";
 import * as region from "@src/layout/region";
 import traverse from "@src/layout/traverse";
 import { checkDocumentStyles } from "@src/layout/style";
+import * as scroll from "src/interaction/scroll";
```
swaathee, line 13: should this be @src

packages/clarity-js/src/interaction/scroll.ts
```diff
@@ -93,6 +100,17 @@ function similar(last: ScrollState, current: ScrollState): boolean {
...
+        dimension.log(Dimension.InitialScrollTop, top?.hash?.join(Constant.Dot));
+    }
+    if (initialBottom) {
+        const bottom = metadata(initialBottom, null);
+        dimension.log(Dimension.InitialScrollBottom, bottom?.hash?.join(Constant.Dot));
+    }
+}
+
 export function stop(): void {
     clearTimeout(timeout);
```
swaathee, line 115: clear initialtop and bottom

packages/clarity-js/src/interaction/encode.ts
```diff
@@ -117,13 +117,17 @@ export default async function (type: Event, ts: number = null): Promise<void> {
         case Event.Scroll:
             for (let entry of scroll.state) {
                 let sTarget = metadata(entry.data.target as Node, entry.event);
+                const top = metadata(entry.data.top as Node, entry.event);
+                const bottom = metadata(entry.data.bottom as Node, entry.event);
+                const sTopHash = top.hash ? top.hash.join(Constant.Dot) : Constant.Empty;
```
swaathee, line 122: ? after top to handle not discovered cases

swaathee (approved):

ender336 (approved):
:shipit:

PR #591, opened 2024-04-29
do not send previous if parent is null

swaathee:


AbdelrhmanMagdy (approved):

packages/clarity-js/src/layout/encode.ts
```diff
@@ -93,8 +93,10 @@ export default async function (type: Event, timer: Timer = null, ts: number = nu
                                     let box = size(value);
                                     let factor = mangle ? -1 : 1;
                                     tokens.push(value.id * factor);
-                                    if (value.parent && active) { tokens.push(value.parent); }
-                                    if (value.previous && active) { tokens.push(value.previous); }
+                                    if (value.parent && active) { 
+                                        tokens.push(value.parent); 
+                                        if (value.previous && active) { tokens.push(value.previous); }
```
ender336, line 98: >active [](http://example.com/codeflow?start=62&length=6)

nit: active is already guaranteed to be true - can just check value.previous here

ender336 (approved):
:shipit:

PR #604, opened 2024-05-14
breaking out the style states into 2 fields and ensuring we dont rese…

ender336:
…t in the middle of capturing subsequent pages

### DIFF
--- a/packages/clarity-js/src/layout/encode.ts
+++ b/packages/clarity-js/src/layout/encode.ts
@@ -38,17 +38,15 @@ export default async function (type: Event, timer: Timer = null, ts: number = nu
             region.reset();
             break;
         case Event.StyleSheetAdoption:
-            for (let entry of style.state) {
+        case Event.StyleSheetUpdate:
+            for (let entry of style.sheetAdoptionState) {
                 tokens = [entry.time, entry.event];
                 tokens.push(entry.data.id);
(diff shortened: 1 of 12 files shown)


packages/clarity-js/src/layout/style.ts
```diff
@@ -7,8 +7,10 @@ import { getId, getNode } from "@src/layout/dom";
 import * as core from "@src/core";
 import { getCssRules } from "./node";
 import * as metric from "@src/data/metric";
+import { schedule } from "@src/core/task";
```
swaathee, line 10: remove

packages/clarity-js/src/layout/encode.ts
```diff
@@ -38,24 +38,23 @@ export default async function (type: Event, timer: Timer = null, ts: number = nu
             region.reset();
             break;
         case Event.StyleSheetAdoption:
-            for (let entry of style.state) {
+        case Event.StyleSheetUpdate:
```
swaathee, line 41: is this intentional that we process both when one happens?

ender336: probably unnecessary - but I thought this was still safer and set us up for the future better

swaathee (approved):

AbdelrhmanMagdy (approved):

PR #607, opened 2024-05-20
Logging lcp target

AbdelrhmanMagdy:


packages/clarity-js/src/data/metric.ts
```diff
@@ -42,6 +42,12 @@ export function max(metric: Metric, value: number): void {
     }
 }
 
+export function logTarget(event: Event, target: number): void {
+    if(event == Event.LargestPaintTarget){
```
ender336, line 46: triple ===

packages/clarity-js/src/performance/observer.ts
```diff
@@ -81,6 +83,10 @@ function process(entries: PerformanceEntryList): void {
                 break;
             case Constant.LCP:
                 if (visible) { metric.max(Metric.LargestPaint, entry.startTime); }
+                const targetNode = dom.get((entry as any).element)?.id 
+                if(targetNode){
```
ender336, line 87: nit, space between if and (

packages/clarity-js/src/performance/observer.ts
```diff
@@ -81,6 +83,10 @@ function process(entries: PerformanceEntryList): void {
                 break;
             case Constant.LCP:
                 if (visible) { metric.max(Metric.LargestPaint, entry.startTime); }
```
ender336, line 85: do we want to only log targetNode  if it is visible? not sure why we log the metric when visible, but the rest of it regardless

ender336 (approved):

PR #611, opened 2024-05-26
Updating our CLS approach

AbdelrhmanMagdy:


packages/clarity-js/src/performance/observer.ts
```diff
@@ -83,9 +88,37 @@ function process(entries: PerformanceEntryList): void {
...
+    // For more info: https://web.dev/articles/cls
+    if (
+        sessionValue &&
+        lastSessionEntry && (entry.startTime - lastSessionEntry.startTime < 1000) &&
+        firstSessionEntry && (entry.startTime - firstSessionEntry.startTime < 5000)
+    ) {
+        sessionValue += entry["value"];
+        sessionEntries.push(entry);
+    } else {
+        sessionValue = entry["value"];
```
ender336, line 110: >sessionValue [](http://example.com/codeflow?start=8&length=12)

we used to multiply by 1000, how are we handling the transition for the backend here?

packages/clarity-js/src/performance/observer.ts
```diff
@@ -83,9 +88,37 @@ function process(entries: PerformanceEntryList): void {
...
+        sessionEntries = [entry];
+    }
+    // Scale the value to avoid sending back floating point number
+    metric.max(Metric.CumulativeLayoutShift, entry["value"] * 1000)
```
ender336, line 114: >entry["value"] [](http://example.com/codeflow?start=45&length=14)

should this maybe be sessionValue instead? As it stands, I'm not sure I understand what the purpose of sessionValue would have been (or any of the updated logic)

ender336 (changes requested):
:clock1:

toby-walker:
Very old PR. Closing.

PR #618, opened 2024-06-14
Conversion config supporting click text reporting

ender336:


swaathee (approved):

packages/clarity-js/src/core/config.ts
```diff
@@ -20,7 +20,8 @@ let config: Config = {
     action: null,
     dob: null,
     delayDom: false,
-    throttleDom: true
+    throttleDom: true,
+    conversions: false
```
keyrun, line 24: Nit: I like adding a comma at the end as it will cause less merge conflicts.

keyrun (approved):

packages/clarity-js/src/core/config.ts
ender336: fair :)

PR #623, opened 2024-06-26
Ensuring newly discovered nodes are appropriately tracked

ender336:
lest we have a dangling reference that can break visualization when node is reused


swaathee (approved):

AbdelrhmanMagdy (approved):

ender336:
found a way of handling this better in visualize, abandoning this one

PR #633, opened 2024-07-11
missing the case where an animation was created directly via element.…

ender336:
…animate

### DIFF
--- a/packages/clarity-js/src/layout/animation.ts
+++ b/packages/clarity-js/src/layout/animation.ts
@@ -31,9 +31,9 @@ export function start(): void {
         if (elementAnimate === null) {
             elementAnimate = Element.prototype.animate;
             Element.prototype.animate = function(): Animation {
-                var retVal = elementAnimate.apply(this, arguments);
-                animateWorker(retVal, "play");
-                return retVal;
+                var createdAnimation = elementAnimate.apply(this, arguments);
+                trackAnimationOperation(createdAnimation, "play");
+                return createdAnimation;
             }
         }
     }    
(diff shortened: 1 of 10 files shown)


packages/clarity-js/src/layout/animation.ts
```diff
@@ -59,43 +68,46 @@ function overrideAnimationHelper(functionToOverride: () => void, name: string) {
...
       }
     }
-  }
\ No newline at end of file
+  }
+
+function animateWorker(animation: Animation, name: string) {
+    if (core.active()) {
+        let effect = <KeyframeEffect>animation.effect;
+        let target = getId((<any>animation.effect).target);
```
keyrun, line 80: ><any>animation.effect [](http://example.com/codeflow?start=28&length=21)

Just use effect here instead of the cast and animation.effect

keyrun (approved):
:shipit:

keyrun (approved):

swaathee (approved):

packages/clarity-js/src/layout/animation.ts
```diff
@@ -27,6 +28,14 @@ export function start(): void {
         overrideAnimationHelper(animationPause, "pause");
         overrideAnimationHelper(animationCancel, "cancel");
         overrideAnimationHelper(animationFinish, "finish");
+        if (elementAnimate === null) {
+            elementAnimate = Element.prototype.animate;
+            Element.prototype.animate = function(): Animation {
+                var createdAnimation = elementAnimate.apply(this, arguments);
```
swaathee, line 34: usually we do our logic before applying (see line 71) - is it intentional to do it the other way?

PR #637, opened 2024-07-24
Improve track flag

MustafaLeithy:


packages/clarity-js/src/data/metadata.ts
```diff
@@ -120,6 +123,9 @@ export function consent(status: boolean = true): void {
   if (core.active()) {
     config.track = true;
     track(user(), BooleanFlag.True);
+    save();
+    read();
+    encode(Event.Consent);
```
ender336, line 128: >encode(Event.Consent); [](http://example.com/codeflow?start=4&length=22)

can we please add the status to the consent event, and log in both positive cases

ender336:
make sure to upgrade the version too please

PR #642, opened 2024-07-29
Adding extra metadata signals

AbdelrhmanMagdy:


packages/clarity-js/src/data/metadata.ts
```diff
@@ -17,6 +17,9 @@ let rootDomain = null;
 export function start(): void {
   rootDomain = null;
   const ua = navigator && "userAgent" in navigator ? navigator.userAgent : Constant.Empty;
+  const timezone = Intl?.DateTimeFormat().resolvedOptions().timeZone ?? '';
```
ender336, line 20: >Intl?.DateTimeFormat().resolvedOptions().timeZone [](http://example.com/codeflow?start=19&length=49)

does look like there are browsers which won't support all of this:

[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#browser_compatibility](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#browser_compatibility)

should do something safer like Intl?.DateTimeFormat?.()?.resolvedOptions?.()?.timeZone

packages/clarity-js/src/data/metadata.ts
```diff
@@ -17,6 +17,9 @@ let rootDomain = null;
...
+  const timezoneOffset = new Date().getTimezoneOffset().toString();
+  const ancestorOrigins = Array.from(window.location.ancestorOrigins).toString();
```
ender336, line 22: this throws in firefox - ancestorOrigins isn't defined there.

packages/clarity-js/src/data/metadata.ts
AbdelrhmanMagdy: ResolvedOptions is actually supported by all browsers, I added the extra safe validations anyways.

ender336 (approved):

alaa-megahed (approved):

PR #651, opened 2024-08-17
Fix metadata callbacks for long/multiple sessions

malaref:
Metadata callbacks with `!wait` flag are executed once and not added to the list of callbacks in case another session starts for the same page.


malaref:

ender336 (approved):

AbdelrhmanMagdy (approved):

PR #662, opened 2024-09-06
adding telemetry callback for clicks being on the wrong element

ender336:


AbdelrhmanMagdy (approved):

packages/clarity-visualize/types/visualize.d.ts
```diff
@@ -34,6 +34,7 @@ export class Visualizer {
 export type ResizeHandler  = (width: number, height: number) => void;
 export type ErrorLogger = (error: Error) => void;
 export type LinkHandler = (link: string, id: string, linkType: string) => string;
+export type ClickLogger = (time: number, x: number, y: number, nodeId: number) => void;
```
keyrun, line 37: >time: number, x: number, y: number, nodeId: number [](http://example.com/codeflow?start=27&length=50)

I wonder if a object with these properties would be better than arguments. Will make it more flexible for change if needed.

keyrun (approved):
:shipit:

PR #679, opened 2024-10-15
Add track signal as dimension

MustafaLeithy:


AbdelrhmanMagdy (approved):

packages/clarity-js/src/data/metadata.ts
```diff
@@ -127,9 +131,11 @@ export function consent(status: boolean = true): void {
     return;
   }
 
+  trackConsent.consent(status);
   if (core.active()) {
     config.track = true;
     track(user(), BooleanFlag.True);
+    save();
```
swaathee, line 138: this calls processcallback which triggers user defined callbacks multiple times if we are on a config.track page but receive consent. we should make sure that callback is only called once.

swaathee (changes requested):

packages/clarity-js/src/data/metadata.ts
MustafaLeithy: Yes we had that done in the previous PR as well, but I missed adding this here. Great catch!

swaathee (approved):

PR #680, opened 2024-10-15
Add id to pointer data to handle multi-touch screens

malaref:
Add touch identifier to pointer data to handle multi-touch screens
Ref: https://www.w3.org/TR/touch-events/#widl-Touch-identifier

### DIFF
--- a/packages/clarity-js/src/interaction/encode.ts
+++ b/packages/clarity-js/src/interaction/encode.ts
@@ -37,6 +37,7 @@ export default async function (type: Event, ts: number = null): Promise<void> {
                     tokens.push(pTarget.id);
                     tokens.push(entry.data.x);
                     tokens.push(entry.data.y);
+                    if (entry.data.id !== undefined) { tokens.push(entry.data.id); }
                     queue(tokens);
                     baseline.track(entry.event, entry.data.x, entry.data.y);
                 }
--- a/packages/clarity-js/src/interaction/pointer.ts
+++ b/packages/clarity-js/src/interaction/pointer.ts
@@ -58,8 +58,13 @@ function touch(event: Event, root: Node, evt: TouchEvent): void {
             x = x && frame ? x + Math.round(frame.offsetLeft) : x;
             y = y && frame ? y + Math.round(frame.offsetTop) : y;
 
+            // identifier is 0-based, unique for each touch point and resets when all fingers are lifted
+            // that is not a part of the spec, but it is how it is implemented in browsers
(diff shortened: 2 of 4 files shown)


packages/clarity-js/src/interaction/encode.ts
```diff
@@ -37,6 +37,7 @@ export default async function (type: Event, ts: number = null): Promise<void> {
                     tokens.push(pTarget.id);
                     tokens.push(entry.data.x);
                     tokens.push(entry.data.y);
+                    if (entry.data.id !== undefined) { tokens.push(entry.data.id); }
```
toby-walker, line 40: When would this be undefined?

packages/clarity-js/src/interaction/pointer.ts
```diff
@@ -57,9 +57,10 @@ function touch(event: Event, root: Node, evt: TouchEvent): void {
             let y = "clientY" in entry ? Math.round(entry["clientY"] + d.scrollTop) : null;
             x = x && frame ? x + Math.round(frame.offsetLeft) : x;
             y = y && frame ? y + Math.round(frame.offsetTop) : y;
+            const id = "identifier" in entry ? entry["identifier"] : undefined;
```
toby-walker, line 60: The server side is assuming id=0 is the primary pointer and the rest are != o. However there is no guarantee in teh spec of that behavior. We need to do the number ourselves and map from the arbitrary touch id to the simple numbering scheme.

Basic idea: 

* Keep a dictionary of id -> touch number
* When it's a touchstart event then add a new entry to the dictionary with the touch # assigned starting from 0.

malaref: The `identifier` is not arbitrary, it's 0-based and it resets when all the fingers are off the screen. It's not a part of the spec, but that's the implemented behaviour in browsers (tested on Chrome-based ones as well as Firefox).

packages/clarity-js/src/interaction/encode.ts
malaref: When it's a mouse (click) event, not touch.

packages/clarity-js/src/interaction/pointer.ts
toby-walker: Can you please add a comment that capture the above (starts at zero and confirmed on the 3 major browse engines?

malaref: done

ender336 (approved):
:shipit:

jjshaws (approved):

PR #681, opened 2024-10-17
Improving INP metric calculation

AbdelrhmanMagdy:


packages/clarity-js/src/performance/interaction.ts
```diff
@@ -0,0 +1,124 @@
...
+const longestInteractionMap: Map<number, Interaction> = new Map();
+
+/**
+ * Update the approx number of interactions estimate count if the interactionCount is not supported.
+ * The difference between `maxKnownInteractionId` and `minKnownInteractionId` gives us a rough range of how many interactions have occurred. 
+ * Dividing by 7 helps approximate the interaction count more accurately, since interaction IDs may not always increase sequentially.
+ */
+const countInteractions = (entry: PerformanceEventTiming) => {
+  if ('interactionCount' in Performance) return;
+
```
ender336, line 26: imo it'd be a bit clearer to have this function put Performance.interactionCount into interactionCountEstimate, then getInteractionCount can just always return interactionCountEstimate || 0 (and you don't have to check for interactionCount in performance twice or cast anything to number

packages/clarity-js/src/performance/interaction.ts
```diff
@@ -0,0 +1,124 @@
...
+  const minLongestInteraction =
+    longestInteractionList[longestInteractionList.length - 1];
+
+  const existingInteraction = longestInteractionMap.get(entry.interactionId!);
+
+  // Either update existing, add new, or replace shortest interaction if necessary
+  if (
+    existingInteraction ||
+    longestInteractionList.length < MAX_INTERACTIONS_TO_CONSIDER ||
+    entry.duration > minLongestInteraction.latency
```
ender336, line 101: >minLongestInteraction [](http://example.com/codeflow?start=21&length=21)

don't you run the risk of a null reference here? On our first execution longestInteractionList was [] so minLongestInteraction will be undefined, which should throw at this point?

ender336 (approved):
:shipit:

malaref (approved):

PR #692, opened 2024-10-30
Enhance Error Reporting with Beacon API and Fallback to XHR

Rajaniraiyn:
This PR improves the error reporting mechanism in Microsoft’s Clarity data collection and analytics package by introducing support for the navigator.sendBeacon API with a fallback to XMLHttpRequest for enhanced reliability and performance. 

Utilizing the navigator.sendBeacon API for asynchronous, non-blocking error reporting to the server, ensuring data is sent effectively even when users navigate away from the page.


Rajaniraiyn:

PR #710, opened 2024-11-18
Throttling instead of suspending mutations

ender336:


AbdelrhmanMagdy (approved):

packages/clarity-js/src/layout/mutation.ts
```diff
@@ -137,41 +140,62 @@ function handle(m: MutationRecord[]): void {
...
   }
+
+  let throttledMutationsToProcess = throttledMutations;
+  throttledMutations = [];
+  let processedMutations = false;
+  while (throttledMutationsToProcess.length > 0) {
+    let throttledMutationToProcess = throttledMutationsToProcess.shift();
+    processMutation(timer, throttledMutationToProcess.mutation, time(), throttledMutationToProcess.timestamp);
+    processedMutations = true;
+  }
```
malaref, line 188: Should this part be at the top of the function?
check previously throttled things -> check new things
instead of
check new things -> check previously throttled things

---

main reason: in case order matters, this may add some new things before throttled things that should've been processed
not-so-main reason: every time something is throttled, we immediately check it again

ender336: I actually want the new things to overwrite the throttled mutations (i improved the logic a bit and made it more clear now thanks for calling this out)

PR #717, opened 2024-11-30
Add primary flag for touch events to handle multi-touch scenarios

nairmanu:


packages/clarity-js/src/interaction/pointer.ts
```diff
@@ -11,6 +11,8 @@ import encode from "./encode";
 
 export let state: PointerState[] = [];
 let timeout: number = null;
+let activeTouchPointId = 0;
+let activeTouchPointIds = new Set<number>();
```
malaref, line 15: nit: this can be `const`

malaref (approved):

packages/clarity-js/src/interaction/pointer.ts
nairmanu: the value changes right? how can it be const?

ender336: const for a set means that the pointer doesn't change essentially, it doens't mean the contents of the set are the same (like an array you can have const x = []; x.push(1); x.push(2); //you just can't then do x = [2,3,4]

packages/clarity-js/src/interaction/pointer.ts
```diff
@@ -58,13 +60,26 @@ function touch(event: Event, root: Node, evt: TouchEvent): void {
...
+                        activeTouchPointId = id;
+                    }
+                    activeTouchPointIds.add(id);
+                    break;
+                case Event.TouchEnd:
+                case Event.TouchCancel:
+                    activeTouchPointIds.delete(id);
+                    break;
+            }
+            const isPrimary: boolean = activeTouchPointId === id;
```
ender336, line 79: nit, you don't need to define the type here, it's implicitly defined by activeTouchPointId === id

packages/clarity-js/src/interaction/pointer.ts
```diff
@@ -58,13 +60,26 @@ function touch(event: Event, root: Node, evt: TouchEvent): void {
...
+
             // Check for null values before processing this event
-            if (x !== null && y !== null) { handler({ time: t, event, data: { target: target(evt), x, y, id } }); }
+            if (x !== null && y !== null) { handler({ time: t, event, data: { target: target(evt), x, y, id, isPrimary } }); }
```
ender336, line 82: don't you need to add something into encode so this value actually gets sent?

packages/clarity-js/src/interaction/pointer.ts
nairmanu: updated

packages/clarity-js/src/interaction/pointer.ts
nairmanu: changed to const

packages/clarity-js/src/interaction/pointer.ts
nairmanu: yes! added them. thanks!

ender336 (approved):

PR #720, opened 2024-12-11
Additional logging to capture long task details

nairmanu:


swaathee (approved):

packages/clarity-js/src/interaction/pointer.ts
```diff
@@ -76,13 +80,19 @@ function touch(event: Event, root: Node, evt: TouchEvent): void {
...
             }
-            const isPrimary = activeTouchPointId === id;
+            const isPrimary = hasPrimaryTouch && primaryTouchId === id;
 
             // Check for null values before processing this event
             if (x !== null && y !== null) { handler({ time: t, event, data: { target: target(evt), x, y, id, isPrimary } }); }
+
+            // Reset primary touch point id once touch event ends
+            if (event === Event.TouchCancel || event === Event.TouchEnd) {
+                if (primaryTouchId === id) { hasPrimaryTouch = false; }
```
swaathee, line 90: reset primarytouchId here

nairmanu: wouldnt matter as the value we reset it to might end up matching the new touch id, hence the use of hasPrimaryTouch boolean

packages/clarity-js/src/core/measure.ts
```diff
@@ -13,6 +14,7 @@ export default function (method: Function): Function {
         if (duration > config.longTask) {
             metric.count(Metric.LongTaskCount);
             metric.max(Metric.ThreadBlockedTime, duration);
+            internal.log(Code.FunctionExecutionTime, Severity.Info, `${method.displayName || method.name}-${duration}`);
```
ender336, line 17: can we give these functions ids rather than full string names (like 1,2,3,etc). we are trying to reduce the size of clarity.js

ender336: also shorten displayName please

nairmanu: done

PR #729, opened 2025-01-01
Send url dimension with all payloads

alaa-megahed:


AbdelrhmanMagdy (approved):

packages/clarity-js/src/data/metadata.ts
```diff
@@ -82,6 +82,11 @@ export function start(): void {
   track(u);
 }
 
+export function recurringMetadata(): void {
+  // Set the metadata values that should be sent with every payload
+  dimension.log(Dimension.Url, scrub.url(location.href, !!electron));
```
ender336, line 87: This is what the envelope is for - let's put the data there instead. Also to save bytes can we remove query params?

alaa-megahed: I'll move it to the envelope. For context, this change is needed to enable calculating some smart events in decode. According to Amira, in some cases they use url parameters. I think we can set some threshold for the length and truncate long URLs to save bytes.

packages/clarity-js/src/core/scrub.ts
```diff
@@ -87,17 +88,22 @@ export function text(value: string, hint: string, privacy: Privacy, mangle: bool
...
 }
 
-export function url(input: string, electron: boolean = false): string {
+export function url(input: string, electron: boolean = false, truncate: boolean = false): string {
+    let result = input;
     // Replace the URL for Electron apps so we don't send back file:/// URL
-    if (electron) { return `${Data.Constant.HTTPS}${Data.Constant.Electron}`; }
+    if (electron) { result = `${Data.Constant.HTTPS}${Data.Constant.Electron}`; }
 
     let drop = config.drop;
```
ender336, line 96: looks like you need to wrap these non electron statements in an else now - we don't want them to execute when it's electron

ender336 (approved):

PR #733, opened 2025-01-08
Throttle mutations during critical period for INP

nairmanu:


ender336 (approved):

packages/clarity-js/src/layout/mutation.ts
```diff
@@ -263,13 +274,14 @@ function names(nodes: NodeList): string {
 async function processNodeList(list: NodeList, source: Source, timer: Timer, timestamp: number): Promise<void> {
   let length = list ? list.length : 0;
   for (let i = 0; i < length; i++) {
+    let node = list[i];
```
malaref, line 277: nit: I don't think this aliasing is needed... and if so, it'd be a `const` to mark it as an alias, as I was looking for somewhere it was re-assigned depending on the newly added flag

packages/clarity-js/src/layout/mutation.ts
```diff
@@ -217,15 +221,19 @@ function cleanHistory(): void {
   }
 }
 
-function track(m: MutationRecord, timer: Timer, instance: number, timestamp: number): string {
+function track(m: MutationRecord, timer: Timer, instance: number, timestamp: number, throttleMutations?: boolean): string {
```
malaref, line 224: nit: why add a new (optional) param? both this and the caller can use `config.throttleMutations` directly and it doesn't change

malaref (approved):

packages/clarity-js/src/layout/mutation.ts
nairmanu: there were two list lookups and hence the alias. agree with using const vs let. changed accordingly.

packages/clarity-js/src/layout/mutation.ts
nairmanu: good point, changed!

PR #745, opened 2025-01-21
Adding configurations for INP experiments

malaref:


packages/clarity-js/src/layout/mutation.ts
```diff
@@ -133,7 +133,7 @@ export function stop(): void {
 
 export function active(): void {
   activePeriod = time() + Setting.MutationActivePeriod;
-  criticalPeriod = time() + Setting.INPCriticalPeriod;
+  criticalPeriod = time() + config.criticalMs;
```
nairmanu, line 136: if criticalMs is not defined, can we read from Setting instead as fallback?

packages/clarity-js/src/layout/mutation.ts
```diff
@@ -233,7 +233,9 @@ function track(m: MutationRecord, timer: Timer, instance: number, timestamp: num
     let element = target && target.selector ? target.selector.join() : m.target.nodeName;
     let parent = value.selector ? value.selector.join() : Constant.Empty;
     // Check if its a low priority (e.g., ads related) element mutation happening during critical period
-    const lowPriMutation = config.throttleMutations && critical && element.indexOf("adthrive-ad") !== -1;
+    const lowPriMutation =
+      config.throttleMutations && critical &&
+      (config.discard.length === 0 || config.discard.some((key) => element.includes(key)));
```
nairmanu, line 238: this means if there are no discard elements explicitly defined, we are throttling all mutations? if so, can u pls add a comment saying so

packages/clarity-js/src/layout/mutation.ts
```diff
@@ -257,8 +259,10 @@ function track(m: MutationRecord, timer: Timer, instance: number, timestamp: num
       if (instance > timestamp + Setting.MutationActivePeriod) {
         return m.type;
       }
-      // we only store the most recent mutation for a given key if it is being throttled
-      throttledMutations[key] = {mutation: m, timestamp};
+      if (!config.dropMutations) {
+        // we only store the most recent mutation for a given key if it is being throttled
+        throttledMutations[key] = {mutation: m, timestamp};
+      }
       return Constant.Throttle;
```
nairmanu, line 266: the return stmt should be within the if condition right? if we are dropping mutation, shoudlnt we return Constant.Empty?

i think we can refactor this to something like

if (config.dropMutations) {
return Constant.Empty;
}
throttledMutations[key] = {mutation: m, timestamp};
...

malaref: I don't think it matters too much, since both `Constant.Throttle` and `Constant.Empty` are treated similarly

packages/clarity-js/src/layout/mutation.ts
malaref: Added

packages/clarity-js/src/layout/mutation.ts
nairmanu: still gives an impression that it is getting throttled and not getting dropped. from a code readability standpoint, isnt it better to use a different constant?

nairmanu (approved):

ender336 (approved):

packages/clarity-js/src/layout/mutation.ts
malaref: Dropping is a part of the overall throttling experiment, and since this is still in the experimentation phase, I generally prefer doing minimal changes to the code flow

PR #755, opened 2025-01-31
Extend baseline to hold pointer move data

nairmanu:
- Added pointer move X, Y coordinates and time to the baseline data structure to track pointer movement during user interactions.
- - Added pointer down X, Y coordinates and time to the baseline data structure to track pointer down during user interactions.
- Added pointer time to the baseline data structure to track the time of pointer movement during user interactions.

### DIFF
--- a/packages/clarity-js/src/data/baseline.ts
+++ b/packages/clarity-js/src/data/baseline.ts
@@ -27,7 +27,14 @@ export function reset(): void {
             pointerX: buffer.pointerX,
             pointerY: buffer.pointerY,
             activityTime: buffer.activityTime,
-            scrollTime: buffer.scrollTime
+            scrollTime: buffer.scrollTime,
+            pointerTime: buffer.pointerTime,
+            moveX: buffer.moveX,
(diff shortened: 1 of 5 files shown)


malaref (approved):

packages/clarity-js/src/data/baseline.ts
```diff
@@ -61,9 +75,20 @@ export function track(event: Event, x: number, y: number, time?: number): void {
             buffer.scrollY = y;
             buffer.scrollTime = time;
             break;
+        case Event.MouseMove:
+            buffer.moveX = x;
+            buffer.moveY = y;
+            buffer.moveTime = time;
+        // intentionally skipping break to ensure pointer data is also updated for mouse move event
```
ender336, line 82: this doesn't seem correct to me. On a mouseMove event we update the downX, downY and downTime?

nairmanu: yes, i was on the fence about this. updated to copy the logic from default to two new blocks

ender336 (approved):

PR #761, opened 2025-02-10
Add flag for click text length

nairmanu:
- Add new field to denote if text is full length or partial in click event
- Updated deprecated `substr` with `substring` method

### DIFF
--- a/packages/clarity-decode/src/interaction.ts
+++ b/packages/clarity-decode/src/interaction.ts
@@ -36,7 +36,8 @@ export function decode(tokens: Data.Token[]): InteractionEvent {
                 link: tokens[11] as string,
                 hash: clickHashes[0],
                 hashBeta: clickHashes.length > 0 ? clickHashes[1] : null,
-                trust: tokens.length > 13 ? tokens[13] as number : Data.BooleanFlag.True
+                trust: tokens.length > 13 ? tokens[13] as number : Data.BooleanFlag.True,
+                isFullText: tokens.length > 14 ? tokens[14] as number : null,
             };
             return { time, event, data: clickData };
         case Data.Event.Clipboard:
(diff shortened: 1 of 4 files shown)


packages/clarity-js/types/interaction.d.ts
```diff
@@ -122,6 +122,12 @@ export interface ClickData {
     link: string;
     hash: string;
     trust: number;
+    isFullText: BooleanFlag;
+}
+
+export interface TextInfo {
```
malaref, line 128: Not sure if this needs to be a dedicated interface in the types as it's only used by one function... Typescript can be infer it from the function directly or it can even be added as a shorthand return type

packages/clarity-decode/src/interaction.ts
```diff
@@ -36,7 +36,8 @@ export function decode(tokens: Data.Token[]): InteractionEvent {
                 link: tokens[11] as string,
                 hash: clickHashes[0],
                 hashBeta: clickHashes.length > 0 ? clickHashes[1] : null,
-                trust: tokens.length > 13 ? tokens[13] as number : Data.BooleanFlag.True
+                trust: tokens.length > 13 ? tokens[13] as number : Data.BooleanFlag.True,
+                isFullText: tokens.length > 14 ? tokens[14] as number : Data.BooleanFlag.False,
```
malaref, line 40: Why is it `False` by default?

packages/clarity-js/src/interaction/click.ts
```diff
@@ -88,19 +90,24 @@ function link(node: Node): HTMLAnchorElement {
     return null;
 }
 
-function text(element: Node): string {
+function text(element: Node): TextInfo {
     let output = null;
+    let t = null;
```
malaref, line 95: I don't think this variable needs to be moved out of its scope

malaref (approved):

AbdelrhmanMagdy (approved):

PR #775, opened 2025-02-24
Fix memory leak issues with mutation observers

nairmanu:
Introduces a new `observedNodes` map in the `mutation.ts` file to ensure duplicate observers are not added to same node.

### DIFF
--- a/packages/clarity-js/src/layout/mutation.ts
+++ b/packages/clarity-js/src/layout/mutation.ts
@@ -35,6 +35,7 @@ let throttleDelay: number = null;
 let activePeriod = null;
 let history: MutationHistory = {};
 let criticalPeriod = null;
+let observedNodes: WeakMap<Node, MutationObserver> = new WeakMap<Node, MutationObserver>();
 
 // We ignore mutations if these attributes are updated
 const IGNORED_ATTRIBUTES = ["data-google-query-id", "data-load-complete", "data-google-container-id"];
@@ -47,6 +48,7 @@ export function start(): void {
   activePeriod = 0;
   history = {};
   criticalPeriod = 0;
+  observedNodes = new WeakMap<Node, MutationObserver>();
 
   // Some popular open source libraries, like styled-components, optimize performance
   // by injecting CSS using insertRule API vs. appending text node. A side effect of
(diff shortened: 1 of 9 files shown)


ender336 (approved):

packages/clarity-js/src/layout/mutation.ts
```diff
@@ -338,11 +364,36 @@ async function processNodeList(list: NodeList, source: Source, timer: Timer, tim
...
   }
 }
 
+function disconnectObserversInIFrame(frame: Node) {
+  if (iframeObserversMap.has(frame)) {
+    const obs = iframeObserversMap.get(frame);
+    obs.forEach((o: MutationObserver) => {
+      o.disconnect();
+
+      if (observedNodes.has(frame)) {
```
malaref, line 384: Shouldn't this ask for the observed document node, not the iframe? If not, it should be taken out of the enclosing `forEach`?

packages/clarity-js/src/layout/mutation.ts
```diff
@@ -47,6 +49,8 @@ export function start(): void {
   activePeriod = 0;
   history = {};
   criticalPeriod = 0;
+  iframeObserversMap = new WeakMap<Node, Set<MutationObserver>>();
+  observedNodes = new WeakMap<Node, MutationObserver>();
```
malaref, line 53: Why is `observedNodes` needed? The only use I can see is the `disconnect` at the top of `observe` below, but that can be alternatively achieved by `iframeObserversMap`.

nairmanu: Observednodes is not just nodes from iframe but all nodes that are observed. We need this one to keep track of nodes that are observed so that if a second attempt to observe the same node is made, we disconnect the previous one to avoid duplicates

malaref: The point still stands though.. all can be tracked by one `WeakMap<Node, Set<MutationObserver>>` object

nairmanu: it cant be. for iframes, the node is the document inside of the iframe. we need to know when iframe is removed to disconnect the observer on the document. that relationship cannot be derived from the suggested weakmap

packages/clarity-js/src/layout/mutation.ts
nairmanu: yes! changed to track nodes instead of observers for iframe so that the observers can be removed from the list as well

swaathee (approved):

PR #783, opened 2025-03-02
Address memory leak issue with event listeners

nairmanu:
Currently, Clarity is not removing event listeners when observed nodes are removed. This leads to a memory leak issue as the event listeners are not properly cleaned up.
- Refactored event binding to use a `Map` for better management of event listeners and added a new `unbind` function to remove event listeners from specific targets.
- Replaced `observers` array with a `Set` in `mutation.ts` for better management of mutation observers. Added `disconnect` function to handle observer disconnections. 
- Added `iframeContentMap` in `dom.ts` to map iframe elements to their content documents and windows. Introduced `iframeContent`, `removeIFrame`, and `ignore` functions to manage iframe-related data and event listeners.

### DIFF
--- a/packages/clarity-js/src/core/event.ts
+++ b/packages/clarity-js/src/core/event.ts
@@ -2,7 +2,7 @@ import { BrowserEvent, Constant } from "@clarity-types/core";
 import api from "./api";
 import measure from "./measure";
 
-let bindings: BrowserEvent[] = [];
+let bindings: Map<EventTarget, BrowserEvent[]> = new Map();
 
 export function bind(target: EventTarget, event: string, listener: EventListener, capture: boolean = false, passive: boolean = true): void {
   listener = measure(listener) as EventListener;
@@ -10,7 +10,11 @@ export function bind(target: EventTarget, event: string, listener: EventListener
   // E.g. Iframe may start off as same-origin but later turn into cross-origin, and the following lines will throw an exception.
   try {
     target[api(Constant.AddEventListener)](event, listener, { capture, passive });
-    bindings.push({ event, target, listener, options: { capture, passive } });
+    if (!has(target)) {
+      bindings.set(target, []);
+    }
+
+    bindings.get(target).push({ event, listener, options: { capture, passive } });
   } catch {
     /* do nothing */
   }
@@ -18,13 +22,32 @@ export function bind(target: EventTarget, event: string, listener: EventListener
 
 export function reset(): void {
   // Walk through existing list of bindings and remove them all
-  for (let binding of bindings) {
+  bindings.forEach((bindingsPerTarget: BrowserEvent[], target: EventTarget) => {
+    resetByTarget(bindingsPerTarget, target);
+  });
+
+  bindings = new Map();
+}
+
+export function unbind(target: EventTarget) {
+  if (!has(target)) {
+    return;
+  }
+  resetByTarget(bindings.get(target), target);
+}
+
+export function has(target: EventTarget): boolean {
+  return bindings.has(target);
+}
+
+function resetByTarget(bindingsPerTarget: BrowserEvent[], target: EventTarget): void {
+  bindingsPerTarget.forEach((binding) => {
     // Wrapping inside try / catch to avoid situations where the element may be destroyed before we get a chance to unbind
     try {
-      binding.target[api(Constant.RemoveEventListener)](binding.event, binding.listener, { capture: binding.options.capture, passive: binding.options.passive });
+      target[api(Constant.RemoveEventListener)](binding.event, binding.listener, { capture: binding.options.capture, passive: binding.options.passive });
     } catch {
       /* do nothing */
     }
-  }
-  bindings = [];
+  });
+  bindings.delete(target);
 }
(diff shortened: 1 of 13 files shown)


ender336 (approved):

swaathee (changes requested):

packages/clarity-js/src/layout/node.ts
```diff
@@ -212,6 +222,20 @@ function observe(root: Node): void {
     interaction.observe(root); // Observe interactions for this root node
 }
 
+function ignore(root: HTMLIFrameElement): void {
+    // iframes will have load event listeners and they should be removed when iframe is removed
+    // from the document
+    unbindTarget(root);
```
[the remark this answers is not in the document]
nairmanu: yes, iframes have load events. so they need to be removed too

packages/clarity-js/src/layout/mutation.ts
```diff
@@ -119,17 +119,13 @@ export function observe(node: Node): void {
...
-      observedNodes.get(node)?.disconnect();
-    }
 
     let m = api(Constant.MutationObserver);
     let observer = m in window ? new window[m](measure(handle) as MutationCallback) : null;
     if (observer) {
       observer.observe(node, { attributes: true, childList: true, characterData: true, subtree: true });
       observedNodes.set(node, observer);
-      observers.push(observer);
+      observers.add(observer);
```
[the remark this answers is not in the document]
nairmanu: yes, since observedNodes is a weakmap, it cannot be iterated on. we keep observers[] when clarity is stopped and we need to remove all

packages/clarity-js/src/layout/node.ts
```diff
@@ -41,11 +42,17 @@ export default function (node: Node, source: Source, timestamp: number): Node {
             dom[call](node, parent, docData, source);
             break;
         case Node.DOCUMENT_NODE:
+            observe(node);
             // We check for regions in the beginning when discovering document and
             // later whenever there are new additions or modifications to DOM (mutations)
-            if (node === document) dom.parse(document);
+            if (node === document) {
```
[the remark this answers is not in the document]
nairmanu: i actually was not sure about this one! its been like this so i didnt want to disturb the current behaviour as part of this change.

packages/clarity-js/src/core/event.ts
```diff
@@ -2,29 +2,52 @@ import { BrowserEvent, Constant } from "@clarity-types/core";
...
+
+export function unbindTarget(target: EventTarget) {
+  if (!bindings.has(target)) {
+    return;
+  }
+  resetByTarget(bindings.get(target), target);
+}
+
+function resetByTarget(bindingsPerTarget: BrowserEvent[], target: EventTarget): void {
+  let isReset = true;
```
malaref, line 40: nit: a better name would be `remove` as `isReset` implies this only happens when this function is called from `reset` which is not the case

packages/clarity-js/src/layout/mutation.ts
```diff
@@ -146,26 +142,36 @@ export function monitor(frame: HTMLIFrameElement): void {
 }
 
 export function stop(): void {
-  for (let observer of observers) {
+  for (let observer of Array.from(observers)) {
```
malaref, line 145: nit: why the additional array construction?

packages/clarity-js/src/layout/node.ts
```diff
@@ -41,11 +42,17 @@ export default function (node: Node, source: Source, timestamp: number): Node {
             dom[call](node, parent, docData, source);
             break;
         case Node.DOCUMENT_NODE:
+            observe(node);
```
malaref, line 45: why is this moved to before styles check?

packages/clarity-js/src/layout/node.ts
```diff
@@ -212,6 +222,20 @@ function observe(root: Node): void {
     interaction.observe(root); // Observe interactions for this root node
 }
 
+function ignore(root: HTMLIFrameElement): void {
```
malaref, line 225: some better name suggestions that would've cut review time and can make further maintenance easier:
* `event.unbindTarget` -> `event.unbind` (it just undoes `event.bind`, and both take `target` as param)
* `mutation.disconnect` -> `mutation.unobserve` (it just undoes `mutation.observe`)
* `ignore` -> `unmonitor` + move it to `mutation.ts` (it [loosely] undoes `mutation.monitor`)

that way, it's obvious why each function is needed to undo something not automatically undone in the `iframe` case and therefore prevents memory leaks

packages/clarity-js/src/layout/node.ts
```diff
@@ -207,11 +213,33 @@ export default function (node: Node, source: Source, timestamp: number): Node {
 }
 
 function observe(root: Node): void {
-    if (dom.has(root)) { return; }
+    if (dom.has(root) || event.has(root)) { return; }
     mutation.observe(root); // Observe mutations for this root node
     interaction.observe(root); // Observe interactions for this root node
 }
 
+export function ignore(root: HTMLIFrameElement): void {
```
swaathee, line 221: nit: can rename to removeObserver

swaathee (approved):

PR #785, opened 2025-03-03
Enhance event tracking logic and improve pointer state comparison

nairmanu:
- Only primary pointer (when available) events should update the baseline state. 
- Pointer “debouncing” should be pointer identifier aware.

### DIFF
--- a/packages/clarity-js/src/interaction/encode.ts
+++ b/packages/clarity-js/src/interaction/encode.ts
@@ -40,12 +40,14 @@ export default async function (type: Event, ts: number = null): Promise<void> {
                     if (entry.data.id !== undefined) { 
                         tokens.push(entry.data.id); 
 
-                        if (entry.data.isPrimary !== undefined) { 
-                            tokens.push(entry.data.isPrimary.toString()); 
+                        if (entry.data.isPrimary !== undefined) {
+                            tokens.push(entry.data.isPrimary.toString());
                         }
                     }
                     queue(tokens);
-                    baseline.track(entry.event, entry.data.x, entry.data.y, entry.time);
+                    if (entry.data.isPrimary === undefined || entry.data.isPrimary) {
+                        baseline.track(entry.event, entry.data.x, entry.data.y, entry.time);
+                    }
                 }
             }
             pointer.reset();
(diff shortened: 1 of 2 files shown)


ender336 (dismissed):

AbdelrhmanMagdy (dismissed):

packages/clarity-js/src/interaction/encode.ts
```diff
@@ -40,12 +40,18 @@ export default async function (type: Event, ts: number = null): Promise<void> {
...
 
-                        if (entry.data.isPrimary !== undefined) { 
-                            tokens.push(entry.data.isPrimary.toString()); 
+                        if (entry.data.isPrimary !== undefined) {
+                            tokens.push(entry.data.isPrimary.toString());
                         }
                     }
                     queue(tokens);
-                    baseline.track(entry.event, entry.data.x, entry.data.y, entry.time);
+                    if (entry.data.isPrimary !== undefined) {
```
malaref, line 48: why not just
```
if (entry.data.isPrimary === undefined || entry.data.isPrimary) {
    baseline.track(entry.event, entry.data.x, entry.data.y, entry.time);
}
```

malaref (approved):

packages/clarity-js/src/interaction/pointer.ts
```diff
@@ -129,7 +129,8 @@ function similar(last: PointerState, current: PointerState): boolean {
     let distance = Math.sqrt(dx * dx + dy * dy);
     let gap = current.time - last.time;
     let match = current.data.target === last.data.target;
-    return current.event === last.event && match && distance < Setting.Distance && gap < Setting.Interval;
+    let sameId = current.data.id !== undefined ? current.data.id === last.data.id : true;
```
swaathee, line 132: this could be tricky, the same element can have multiple ids, think a searchbox with a div inside set up for margin. do you know how much more data we get with this additional check ?

nairmanu: This isn't element id. This is touch id to differentiate between primary vs secondary touches

packages/clarity-js/src/interaction/encode.ts
nairmanu: yes! done

swaathee (approved):

malaref (approved):

PR #801, opened 2025-03-25
Change click reaction for select elements

shawj0:
This change makes reaction false for select elements


shawj0:

packages/clarity-js/src/interaction/selection.ts
```diff
@@ -66,3 +69,19 @@ export function stop(): void {
...
+
+function trackDropdownChanges(root: Node): void {
+    let queryableRoot = root as Document | Element;
+
+    if (queryableRoot.querySelectorAll) {
+        const dropdowns = queryableRoot.querySelectorAll("select");
+
+        dropdowns.forEach(dropdown => {
+            const selectElement = dropdown as HTMLSelectElement;
+            bind(selectElement, "click", () => summary.track(Event.Change, time()), true);
```
swaathee, line 81: can you not track this as a part of the click event itself?  we do get the target element for both click and change events - so ideally if the click / change event happens on a select element - you can add it to summary instead of here.

swaathee (approved):

Niyibitanga (approved):

PR #802, opened 2025-03-27
Add additional info to Click event

nairmanu:
Addition of `tag`, `class`, and `id` properties to click events.

### DIFF
--- a/packages/clarity-decode/src/interaction.ts
+++ b/packages/clarity-decode/src/interaction.ts
@@ -38,6 +38,9 @@ export function decode(tokens: Data.Token[]): InteractionEvent {
                 hashBeta: clickHashes.length > 0 ? clickHashes[1] : null,
                 trust: tokens.length > 13 ? tokens[13] as number : Data.BooleanFlag.True,
                 isFullText: tokens.length > 14 ? tokens[14] as number : null,
+                tag: tokens.length > 15 ? tokens[15] as string : null,
+                class: tokens.length > 16 ? tokens[16] as string : null,
+                id: tokens.length > 17 ? tokens[17] as string : null,
             };
             return { time, event, data: clickData };
         case Data.Event.Clipboard:
--- a/packages/clarity-js/src/interaction/click.ts
+++ b/packages/clarity-js/src/interaction/click.ts
@@ -71,6 +71,9 @@ function handler(event: Event, root: Node, evt: MouseEvent): void {
                 hash: null,
                 trust: evt.isTrusted ? BooleanFlag.True : BooleanFlag.False,
                 isFullText: textInfo.isFullText,
+                tag: tag(t).substring(0, Setting.ClickTag),
+                class: className(t).substring(0, Setting.ClickClass),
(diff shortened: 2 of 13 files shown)


packages/clarity-js/src/interaction/click.ts
```diff
@@ -71,6 +71,9 @@ function handler(event: Event, root: Node, evt: MouseEvent): void {
                 hash: null,
                 trust: evt.isTrusted ? BooleanFlag.True : BooleanFlag.False,
                 isFullText: textInfo.isFullText,
+                tag: tag(t).substring(0, 10),
+                class: className(t).substring(0, 50),
+                id: id(t).substring(0, 25),
```
malaref, line 76: Can we have the limits clear somewhere instead of using magic numbers?

malaref (dismissed):

packages/clarity-js/src/interaction/click.ts
nairmanu: hardcoded the numbers initially as it was meant to be an "experiment". Updated to use constants so that it is ready to merge if everything works fine

packages/clarity-decode/src/interaction.ts
```diff
@@ -38,6 +38,9 @@ export function decode(tokens: Data.Token[]): InteractionEvent {
                 hashBeta: clickHashes.length > 0 ? clickHashes[1] : null,
                 trust: tokens.length > 13 ? tokens[13] as number : Data.BooleanFlag.True,
                 isFullText: tokens.length > 14 ? tokens[14] as number : null,
+                tag: tokens.length > 15 ? tokens[15] as string : null,
```
Niyibitanga, line 41: Is there a reason we use null instead of empty values. I'm thinking it might be safer to avoid cases where the exceptions aren't handled.

Niyibitanga (approved):

packages/clarity-js/src/interaction/click.ts
```diff
@@ -110,15 +113,44 @@ function text(element: Node): TextInfo {
...
-        if (UserInputTags.indexOf(tag) >= 0) {
-            return BooleanFlag.False;
-        }
+    const tagName = tag(element);
+
+    if (UserInputTags.indexOf(tagName) >= 0) {
+        return BooleanFlag.False;
     }
     return BooleanFlag.True;
 }
```
malaref, line 123: refactoring suggestion: I think all three new functions can be written as one parameterized function
```ts
function get(element: Node, attribute: string): string {
    return element.nodeType === Node.ELEMENT_NODE ? (element as HTMLElement)[attribute].toLowerCase() : "";
}
```

or even
```ts
function get(element: Node, attribute: string): string {
    return element[attribute]?.toLowerCase() ?? "";
}
```

packages/clarity-decode/src/interaction.ts
nairmanu: null is used to denote clearly we dont have the value. empty can be confusing - is the value empty or value not present

PR #810, opened 2025-04-10
Stop collecting playback data after limit is reached in lean mode

malaref:
In lean mode, Clarity keeps collecting playback data and keeps it in memory. When it reaches a limit, currently set at 10MB, it shuts down completely. Even `upgrade` signal is ignored afterwards. This changes the behavior to pause collecting playback data before the limit is reached so that other telemetry keeps getting sent and `upgrade` will still be supported.

_Limitation: mutations happening during the pause will be lost._

### DIFF
--- a/packages/clarity-js/src/data/upload.ts
+++ b/packages/clarity-js/src/data/upload.ts
@@ -1,5 +1,5 @@
 import { UploadCallback } from "@clarity-types/core";
-import { BooleanFlag, Check, Constant, EncodedPayload, Event, Metric, Setting, Token, Transit, UploadData, XMLReadyState } from "@clarity-types/data";
+import { BooleanFlag, Check, Code, Constant, EncodedPayload, Event, Metric, Setting, Severity, Token, Transit, UploadData, XMLReadyState } from "@clarity-types/data";
 import * as clarity from "@src/clarity";
 import config from "@src/core/config";
 import measure from "@src/core/measure";
@@ -13,6 +13,7 @@ import * as limit from "@src/data/limit";
 import * as metadata from "@src/data/metadata";
 import * as metric from "@src/data/metric";
 import * as ping from "@src/data/ping";
+import * as internal from "@src/diagnostic/internal";
 import * as timeline from "@src/interaction/timeline";
 import * as region from "@src/layout/region";
 import * as extract from "@src/data/extract";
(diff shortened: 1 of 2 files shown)


packages/clarity-js/src/data/upload.ts
```diff
@@ -47,14 +50,23 @@ export function queue(tokens: Token[], transmit: boolean = true): void {
         let type = tokens.length > 1 ? tokens[1] : null;
         let event = JSON.stringify(tokens);
 
+        if (!config.lean) {
```
nairmanu, line 53: suggestion: its more readable when the conditional block is if true, else

if (config.lean && !leanLimit && playbackBytes + event.length > Setting.PlaybackBytesLimit) {
}
else {
}

packages/clarity-js/src/data/upload.ts
```diff
@@ -28,12 +29,14 @@ let timeout: number = null;
...
 let active: boolean;
 let queuedTime: number = 0;
+let leanLimit = false;
 export let track: UploadData;
 
 export function start(): void {
     active = true;
     discoverBytes = 0;
     playbackBytes = 0;
+    leanLimit = false;
```
nairmanu, line 39: can we call this hasExceededPlaybackLimit or something else that is a bit more meaningful

nairmanu (approved):

packages/clarity-js/src/data/upload.ts
malaref: That will risk `leanLimit` flipping if the next `event.length` is small enough. Instead, it should only flip with `config.lean`

packages/clarity-js/src/data/upload.ts
malaref: With every other variable name abbreviated, that will stand out (in a bad way 😅)

packages/clarity-js/src/data/upload.ts
```diff
@@ -47,14 +50,23 @@ export function queue(tokens: Token[], transmit: boolean = true): void {
         let type = tokens.length > 1 ? tokens[1] : null;
         let event = JSON.stringify(tokens);
 
+        if (!config.lean) {
+            leanLimit = false;
```
ender336, line 54: can you clarify why we need to check the config for every queue call? Shouldn't the config only need to be checked on initial wireup and not every payload?

malaref: Because the `upgrade` signal may happen at any time

ender336 (approved):

PR #814, opened 2025-04-14
version bump for handling dynamic css operations on iframes

ender336:


AbdelrhmanMagdy (approved):
LGTM

malaref (approved):
suggestion: maybe we add a strongly-typed interface extending `Window` instead of using `any`... that would make where and why we do `win.clarityOverrides` clearer

zoeyang7633 (approved):

PR #817, opened 2025-04-17
Update Consent Api

Niyibitanga:
Updating consent API to handle multiple options.


packages/clarity-js/src/data/consent.ts
```diff
@@ -1,5 +1,16 @@
-import { Dimension } from "@clarity-types/data";
+import { Dimension, Event } from "@clarity-types/data";
 import * as dimension from "@src/data/dimension";
+import encode from "./encode";
+
+export let data: string = "";
```
swaathee, line 5: this should be a consent object not a json string - and encode different fields of the object

packages/clarity-js/src/data/consent.ts
```diff
@@ -18,4 +29,16 @@ export function consent(): void {
...
 function trackConsent(consent: ConsentType): void {
     dimension.log(Dimension.Consent, consent.toString());
+}
+
+export function consentv2(status: string): void {
+    data = status;
+}
+
+export function compute(): void {
+    encode(Event.Consent);
```
swaathee, line 39: you should encode consent as soon as you recieve it, it doesnt have to be part of compute

packages/clarity-js/src/data/metadata.ts
```diff
@@ -122,20 +121,34 @@ export function id(): string {
 }
 
 export function consent(status: boolean = true): void {
-  if (!status) {
+  if (status) {
+    consentv2({ adStorage: true, analyticsStorage: true });
+    trackConsent.consent();
```
swaathee, line 126: add a todo to remove this once consentV1 gets deprecated

packages/clarity-js/src/data/metadata.ts
```diff
@@ -122,20 +121,34 @@ export function id(): string {
...
+    return;
+  }
+  consentv2({ adStorage: false, analyticsStorage: false });
+}
+
+export function consentv2(status: Status = {}): void {
```
swaathee, line 132: use typed objects

packages/clarity-js/src/data/consent.ts
```diff
@@ -18,4 +29,16 @@ export function consent(): void {
 
 function trackConsent(consent: ConsentType): void {
```
swaathee, line 30: pass implicit consent from config function to the consentV2 event

swaathee (changes requested):

packages/clarity-js/src/data/consent.ts
Niyibitanga: added

packages/clarity-js/src/data/consent.ts
Niyibitanga: changed

packages/clarity-js/src/data/metadata.ts
```diff
@@ -131,11 +153,13 @@ export function consent(status: boolean = true): void {
     return;
   }
 
-  if (core.active()) {
-    config.track = true;
-    track(user(), BooleanFlag.True);
-    save();
-    trackConsent.consent();
+  if (!consentStatus.ad_Storage || core.active()) {
```
ender336, line 156: this parent clause is kind of strange, I think it's cleaner just in two ifs:

if (!consentStatus.ad_Storage) {
   trackConsent.consentv2(consentStatus);
   }
   
   
   if (core.active()) {
   // other stuff
   }

ender336: I'm also a little confused as to why trackConsent is called when adStorage is true but core isn't active. Don't we probably still want to track that?

packages/clarity-js/types/data.d.ts
```diff
@@ -352,6 +357,14 @@ export const enum XMLReadyState {
     Done = 4
 }
 
+export const enum ConsentSource{
+    Unknown = 0,
+    Implicit = 1,
+    APIsourced = 2,
```
ender336, line 363: nit, just API. (source is already our enum name)

packages/clarity-js/src/data/metadata.ts
Niyibitanga: Assuming you mean why trackConsent **isn't** called when adStorage is true but core isn't active? My understanding is If core isn't active, then Clarity has stopped and there is no data being collected. @swaathee can you confirm?

packages/clarity-js/src/data/metadata.ts
```diff
@@ -132,13 +165,22 @@ export function consent(status: boolean = true): void {
...
     config.track = true;
     track(user(), BooleanFlag.True);
     save();
-    trackConsent.consent();
   }
 }
 
+function getConsent(status: Status): ConsentData {
+  let consent: ConsentData = {
+    ad_Storage: status.ad_Storage === Constant.Granted ? BooleanFlag.True : BooleanFlag.False,
```
swaathee, line 177: will consentStatus be case sensitive? what happens when you pass in Granted instead of granted

packages/clarity-js/types/data.d.ts
```diff
@@ -509,4 +522,15 @@ export interface PerformanceEventTiming extends PerformanceEntry {
...
     latency: number;
+}
+
+export interface Status {
+    ad_Storage?: string;
+    analytics_Storage?: string;
+}
+
+export interface ConsentData {
+    source?: ConsentSource;
```
swaathee, line 533: are these optional fields - why do we need ?

packages/clarity-js/src/data/metadata.ts
```diff
@@ -121,8 +124,38 @@ export function id(): string {
...
   if (!status) {
+    consentv2();
+    return;
+  }
+  
+  consentv2({ ad_Storage: Constant.Granted, analytics_Storage: Constant.Granted });
+  trackConsent.consent();
+}
+
+export function consentv2(latestStatus?: Status, source: number = ConsentSource.API, cb: ConsentCallback = null): void {
```
swaathee, line 138: can you rename latestStatus in api

packages/clarity-js/types/data.d.ts
```diff
@@ -509,4 +522,15 @@ export interface PerformanceEventTiming extends PerformanceEntry {
 export interface Interaction {
     id: number;
     latency: number;
+}
+
+export interface Status {
```
swaathee, line 527: rename status - can we call it consentState instead of status

swaathee (changes requested):

packages/clarity-js/types/data.d.ts
Niyibitanga: no longer optional as they'll always have values

packages/clarity-js/src/data/metadata.ts
```diff
@@ -1,6 +1,6 @@
 import { Time } from "@clarity-types/core";
-import { BooleanFlag, Constant, Dimension, Metadata, MetadataCallback, MetadataCallbackOptions, Metric, Session, User, Setting } from "@clarity-types/data";
 import * as clarity from "@src/clarity";
+import { BooleanFlag, Constant, Dimension, Metadata, MetadataCallback, MetadataCallbackOptions, Metric, Session, User, Setting, consentState, ConsentSource, ConsentData } from "@clarity-types/data";
```
nairmanu, line 3: check consentState

packages/clarity-js/src/data/metadata.ts
```diff
@@ -97,13 +103,19 @@ function userAgentData(): void {
...
   data = null;
   callbacks.forEach(cb => { cb.called = false; });
 }
 
-export function metadata(cb: MetadataCallback, wait: boolean = true, recall: boolean = false): void {
+export function metadata(cb: MetadataCallback, wait: boolean = true, recall: boolean = false, includeConsent: boolean = false): void {
   let upgraded = config.lean ? BooleanFlag.False : BooleanFlag.True;
   let called = false;
+
+  if(includeConsent) {
```
nairmanu, line 115: lint

packages/clarity-js/types/data.d.ts
```diff
@@ -372,6 +382,7 @@ export interface Metadata {
     userId: string;
     sessionId: string;
     pageNum: number;
+    consent?: consentState;
```
nairmanu, line 385: why is this optional

Niyibitanga: Following the existing pattern, avoiding any potential failures if for some reason consent is not included.

packages/clarity-js/src/data/metadata.ts
```diff
@@ -137,30 +154,67 @@ export function metadata(cb: MetadataCallback, wait = true, recall = false): voi
...
-        config.track = false;
-        setCookie(Constant.SessionKey, Constant.Empty, -Number.MAX_VALUE);
-        setCookie(Constant.CookieKey, Constant.Empty, -Number.MAX_VALUE);
-        clarity.stop();
-        window.setTimeout(clarity.start, Setting.RestartDelay);
-        return;
-    }
+  if (!status) {
+    consentv2();
+    return;
```
swaathee, line 135: this should still continue to delete cookie.

packages/clarity-js/src/data/metadata.ts
```diff
@@ -198,7 +252,7 @@ function processCallback(upgrade: BooleanFlag) {
     if (callbacks.length > 0) {
         for (let i = 0; i < callbacks.length; i++) {
             const cb = callbacks[i];
-            if (cb.callback && !cb.called && (!cb.wait || upgrade)) {
+            if (cb.callback && (!cb.called || cb.additionalInfo) && (!cb.wait || upgrade)) {
```
swaathee: can you add a check to see if additionalInfo and consent is updated then do the callback - otherwise, you might callback multiple times

packages/clarity-js/src/data/metadata.ts
```diff
@@ -120,14 +131,20 @@ function userAgentData(): void {
...
     rootDomain = null;
     data = null;
+    consentStatus = null;
     for (const cb of callbacks) {
         cb.called = false;
     }
 }
 
-export function metadata(cb: MetadataCallback, wait = true, recall = false): void {
+export function metadata(cb: MetadataCallback, wait = true, recall = false, additionalInfo = false): void {
```
swaathee: prefer consentInfo to additional

packages/clarity-js/src/data/index.ts
```diff
@@ -47,4 +48,5 @@ export function compute(): void {
     summary.compute();
     limit.compute();
     extract.compute();
+    consent.compute();
```
swaathee, line 47: encode consent every time you get a value instead of doing this at compute - you might miss intermediate states.

packages/clarity-js/src/data/metadata.ts
Niyibitanga: That's handled by consentV2();

packages/clarity-js/src/data/metadata.ts
Niyibitanga: added.

packages/clarity-js/src/data/index.ts
Niyibitanga: We can't encode right away because metadata starts before upload when Clarity starts, i.e consent won't be sent in the payload until the API is called. But I understand the concern of missing intermediate states-so I've refactored to always encode except for the first instance when we get consent using config. That won't be sent right away and will instead be encoded in the compute function when upload has started.

packages/clarity-js/src/data/metadata.ts
```diff
@@ -120,14 +131,22 @@ function userAgentData(): void {
 export function stop(): void {
     rootDomain = null;
     data = null;
+    consentStatus = null;
     for (const cb of callbacks) {
-        cb.called = false;
+        cb.called = cb.consentInfo;
```
swaathee: why change this? at stop, we should set this to false.

packages/clarity-js/src/data/metadata.ts
```diff
@@ -120,14 +131,22 @@ function userAgentData(): void {
...
 
-export function metadata(cb: MetadataCallback, wait = true, recall = false): void {
+export function metadata(cb: MetadataCallback, wait = true, recall = false, consentInfo = false): void {
     const upgraded = config.lean ? BooleanFlag.False : BooleanFlag.True;
     let called = false;
+  
+    if (consentInfo) {
+      data.consent = consentStatus
+    } else{
+        delete data.consent;
```
swaathee: you will have multiple callbacks, some with consentInfo and some without - this will not work. as you return you might want to add consentstatus instead of adding to data

packages/clarity-js/src/data/consent.ts
```diff
@@ -1,14 +1,19 @@
...
     None = 0,
     Implicit = 1,
     General = 2,
 }
 
-export function config(track: boolean): void {
-    trackConsent(track ? ConsentType.Implicit : ConsentType.None);
+export function config(consent: ConsentData): void {
+    trackConsent(consent.analytics_Storage ? ConsentType.Implicit : ConsentType.None);
+    consentv2(consent, false);
```
swaathee: instead of an additional parameter, you can just set data = consent here and not call consentV2 which would avoid the param and if check

packages/clarity-js/src/data/index.ts
swaathee: thanks, added some comment on the refactoring

packages/clarity-js/src/data/metadata.ts
Niyibitanga: At stop and cb.called set to false, all callbacks are executed. Do we want all the callbacks with consentInfo to be executed as well? If a callback does not have consentInfo, cb.called will still be false like always and execute when clarity stops, otherwise it'll be true and not execute.

packages/clarity-js/src/data/metadata.ts
```diff
@@ -120,47 +131,86 @@ function userAgentData(): void {
...
-        trackConsent.consent();
-    }
+export function consentv2(consentState: ConsentState = defaultStatus, source: number = ConsentSource.API): void {
+  consentStatus = {
+    ad_Storage: normalizeConsent(consentState.ad_Storage),
+    analytics_Storage: normalizeConsent(consentState.analytics_Storage)
+  };
+
+  //update consent metadata IFF previously included
+  if(data.consent){
```
swaathee: is this still required?

packages/clarity-js/src/data/metadata.ts
```diff
@@ -194,13 +244,16 @@ export function save(): void {
...
+function processCallback(upgrade: BooleanFlag, consentUpdate: boolean = false): void {
     if (callbacks.length > 0) {
         for (let i = 0; i < callbacks.length; i++) {
             const cb = callbacks[i];
-            if (cb.callback && !cb.called && (!cb.wait || upgrade)) {
-                cb.callback(data, !config.lean);
+            if (cb.callback && (!cb.called || (cb.consentInfo && consentUpdate)) && (!cb.wait || upgrade)) {
+                cb.callback(data, !config.lean, cb.consentInfo ? consentStatus : undefined);
                 cb.called = true;
+                if(consentUpdate){
```
swaathee: can we not do !cb.call || consentUpdate below? - what if we have multiple callback for consent.

packages/clarity-js/src/data/metadata.ts
Niyibitanga: no, removed.

nairmanu (approved):

PR #822, opened 2025-04-29
Observing iframe events in insights script

malaref:


nairmanu (approved):

packages/clarity-js/src/insight/snapshot.ts
```diff
@@ -14,6 +14,16 @@ export function start(): void {
     doc.start();
     getId(document.documentElement); // Pre-discover ID for page root
     interaction.observe(document);
+
+    // observe other document nodes
+    const iframes = document.querySelectorAll("iframe");
+    for (let i = 0; i < iframes.length; i++) {
+        const iframe = iframes[i];
+        if (iframe.contentDocument) {
```
nairmanu, line 22: this can throw exception if iframe is cross origin so better to wrap it in try catch to avoid console errors

packages/clarity-js/src/insight/snapshot.ts
```diff
@@ -14,6 +14,16 @@ export function start(): void {
...
+            getId(iframe.contentDocument.documentElement); // Pre-discover ID for iframe root
```
nairmanu, line 23: do we need to call getId?

nairmanu (changes requested):
pls check for potential errors esp in safari when accessing iframe.contentDocument

PR #827, opened 2025-05-01
Add throttling for scroll, increase interval for similar events (0.8.11)

nairmanu:
This pull request introduces a throttling mechanism to improve the performance of scroll event handling in the `clarity-js` package. The most important changes include the addition of a reusable `throttle` utility function, its integration into the scroll event observer, and the replacement of the unthrottled `recompute` function with a throttled version.

### DIFF
--- a/packages/clarity-js/src/core/throttle.ts
+++ b/packages/clarity-js/src/core/throttle.ts
@@ -1,0 +1,35 @@
+/**
+ * Creates a throttled version of the provided function that only executes at most once
+ * every specified duration in milliseconds, ensuring the last event is not lost.
+ * @param func - The function to throttle.
+ * @param duration - The duration in milliseconds to wait before allowing the next execution.
+ * @returns A throttled version of the provided function.
+ */
+export default function throttle<T extends (...args: any[]) => void>(func: T, duration: number): T {
+  let lastExecutionTime = 0;
+  let timeoutId: ReturnType<typeof setTimeout> | null = null;
+  let lastArgs: Parameters<T> | null = null;
+
+  return function (...args: Parameters<T>) {
+    const now = performance.now();
+    const timeSinceLastExecution = now - lastExecutionTime;
+
+    // If the function is called during the throttling period, store the arguments to ensure we don't drop the last event
+    if (lastExecutionTime !== 0 && timeSinceLastExecution < duration) {
+      lastArgs = args;
+
+      if (timeoutId) return;
+
+      timeoutId = setTimeout(() => {
+        lastExecutionTime = performance.now();
+        func.apply(this, lastArgs!);
+        lastArgs = null;
+        timeoutId = null;
+      }, duration - timeSinceLastExecution);
+    } else {
+      // Execute immediately if outside the throttling period (including the first run)
+      lastExecutionTime = now;
+      func.apply(this, args);
+    }
+  } as T;
+}
--- a/packages/clarity-js/src/interaction/scroll.ts
+++ b/packages/clarity-js/src/interaction/scroll.ts
@@ -5,6 +5,7 @@ import { bind } from "@src/core/event";
 import { schedule } from "@src/core/task";
 import { time } from "@src/core/time";
 import { clearTimeout, setTimeout } from "@src/core/timeout";
+import throttle from "@src/core/throttle";
 import * as dimension from "@src/data/dimension";
 import { iframe } from "@src/layout/dom";
 import { metadata, target } from "@src/layout/target";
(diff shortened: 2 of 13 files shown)


packages/clarity-js/src/core/throttle.ts
```diff
@@ -0,0 +1,13 @@
+export default function throttle<T extends (...args: any[]) => void>(func: T, duration: number): T {
+  let lastExecutionTime = 0;
+
+  return function (...args: Parameters<T>) {
+    const now = performance.now();
+
+    // Execute immediaately for first instance, or after specified duration has elapsed
```
ender336, line 7: nit: immediately

ender336 (approved):

malaref (approved):
I think debouncing can be better than throttling though

packages/clarity-js/src/core/throttle.ts
```diff
@@ -0,0 +1,13 @@
+export default function throttle<T extends (...args: any[]) => void>(func: T, duration: number): T {
+  let lastExecutionTime = 0;
+
+  return function (...args: Parameters<T>) {
+    const now = performance.now();
+
+    // Execute immediately for first instance, or after specified duration has elapsed
+    if (lastExecutionTime === 0 || now - lastExecutionTime >= duration) {
```
toby-walker: This is only one-time and no setTimeout? #Closed

packages/clarity-js/src/interaction/scroll.ts
```diff
@@ -71,6 +72,10 @@ function recompute(event: UIEvent = null): void {
     timeout = setTimeout(process, Setting.LookAhead, Event.Scroll);
 }
 
+const throttledRecompute = throttle((event: UIEvent = null) => {
+    recompute(event);
```
toby-walker, line 76: How does this get executed? What if multiple events are called before the delay time?

packages/clarity-js/src/core/throttle.ts
```diff
@@ -0,0 +1,35 @@
...
+  let lastExecutionTime = 0;
+  let timeoutId: ReturnType<typeof setTimeout> | null = null;
+  let lastArgs: Parameters<T> | null = null;
+
+  return function (...args: Parameters<T>) {
+    const now = performance.now();
+    const timeSinceLastExecution = now - lastExecutionTime;
+
+    // If the function is called during the throttling period, store the arguments to ensure we don't drop the last event
+    if (lastExecutionTime !==0 && timeSinceLastExecution < duration) {
```
malaref: nit: a simpler implementation without nesting `if`s
```ts
    // Always use the latest arguments
    lastArgs = args;

    // Execute immediately if first run or the throttling period has passed
    if (lastExecutionTime == 0 || timeSinceLastExecution > duration) {
      lastExecutionTime = now;
      func.apply(this, lastArgs);
    } else if (!timeoutId) {
        // Otherwise, if no throttled call exists, schedule one
        timeoutId = setTimeout(() => {
          lastExecutionTime = performance.now();
          func.apply(this, lastArgs!);
          timeoutId = null;
        }, duration - timeSinceLastExecution);
      }
    }
``` #Closed

File: packages/clarity-js/src/interaction/scroll.ts [hunk 5]
```diff
@@ -71,6 +72,10 @@ function recompute(event: UIEvent = null): void {
     timeout = setTimeout(process, Setting.LookAhead, Event.Scroll);
 }
 
+const throttledRecompute = throttle((event: UIEvent = null) => {
```
malaref, line 75: nit: no need for an extra inline function
```ts
const throttledRecompute = throttle(recompute, Setting.Throttle);
```

malaref (approved):

packages/clarity-js/src/core/throttle.ts
nairmanu: avoided the nested if in a different way

nairmanu:
> I think debouncing can be better than throttling though

when a user continually scrolls, the function might never run until there’s a sufficient pause, which is problematic as we prefer regular updates.

Niyibitanga (approved):

packages/clarity-js/src/core/throttle.ts
```diff
@@ -0,0 +1,35 @@
+/**
+ * Creates a throttled version of the provided function that only executes at most once
+ * every specified duration in milliseconds, ensuring the last event is not lost.
+ * @param func - The function to throttle.
+ * @param duration - The duration in milliseconds to wait before allowing the next execution.
```
Niyibitanga, line 5: would "delay" be a better param name instead of "duration" for readability?

toby-walker (approved):
:shipit:

PR #842, opened 2025-05-19
Utilizing biome for linting

ender336:


packages/clarity-js/src/data/consent.ts
```diff
@@ -1,10 +1,10 @@
 import { Dimension } from "@clarity-types/data";
 import * as dimension from "@src/data/dimension";
 
-const enum ConsentType {
+enum ConsentType {
```
keyrun, line 4: You want to make this const again?

ender336: thanks, yes

keyrun (approved):
:shipit:

malaref (approved):
This PR is extremely satisfying – thank you!

malaref (approved):

packages/clarity-js/src/core/scrub.ts
```diff
@@ -143,10 +147,12 @@ function regex(): void {
     // Reference: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Unicode_Property_Escapes
     if (unicodeRegex && digitRegex === null) {
         try {
-            digitRegex = new RegExp("\\p{N}", "gu");
-            letterRegex = new RegExp("\\p{L}", "gu");
-            currencyRegex = new RegExp("\\p{Sc}", "gu");
-        } catch { unicodeRegex = false; }
+            digitRegex = /\p{N}/gu;
```
nairmanu, line 150: when using a regex literal instead of a constructor the script will fail to load entirely if Unicode property escapes are not supported, rather than falling back gracefully - try-catch wont work

packages/clarity-js/src/data/custom.ts
```diff
@@ -7,17 +7,12 @@ export let data: CustomData = null;
...
-        a.length < 255
-        ) {
-            if (b && typeof b === Constant.String && b.length < 255) {
-                data = { key: a, value: b};
-            } else {
-                data = { value: a }
-            }
-            encode(Event.Custom);
-        
+    if (core.active() && a && typeof a === "string" && a.length < 255) {
```
nairmanu, line 10: curious, why did we change from Constant.String to "string" ?

PR #846, opened 2025-05-27
viewport null check

zhuangya:
the `rootBounds` of `IntersectionObserverEntry` is possible to be `null` (ref: https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserverEntry/rootBounds), so we need to check if it's `null` before using it

I've already seen a lot of such errors around, so I think it's a good idea to add the check here.


malaref (approved):

zhuangya:
rebased and conflict resolved, requesting for re-request review :)

PR #848, opened 2025-06-05
Log to capture Angular + Zone usage

SeenaAugusty:
Logging to capture when Angular + Zone is used in the Websites.

### DIFF
--- a/packages/clarity-js/src/core/api.ts
+++ b/packages/clarity-js/src/core/api.ts
@@ -6,6 +6,9 @@ export default function api(method: string): string {
     // Zone.js, a popular package for Angular, overrides native browser APIs which can lead to inconsistent state for single page applications.
     // Example issue: https://github.com/angular/angular/issues/31712
     // As a work around, we ensuring Clarity access APIs outside of Zone (and use native implementation instead)
-    internal.log(Code.AngularZone, Severity.Info);
-    return window[Constant.Zone] && Constant.Symbol in window[Constant.Zone] ? window[Constant.Zone][Constant.Symbol](method) : method;
+    const isZone = window[Constant.Zone] && Constant.Symbol in window[Constant.Zone];
+    if (isZone) {
+        internal.log(Code.AngularZone, Severity.Info);
+    }    
+    return  isZone ? window[Constant.Zone][Constant.Symbol](method) : method;
 }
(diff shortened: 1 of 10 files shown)


packages/clarity-js/src/core/api.ts
```diff
@@ -1,8 +1,11 @@
 import { Constant } from "@clarity-types/core";
+import { Code, Severity } from "@clarity-types/data";
+import * as internal from "@src/diagnostic/internal";
 
 export default function api(method: string): string {
     // Zone.js, a popular package for Angular, overrides native browser APIs which can lead to inconsistent state for single page applications.
     // Example issue: https://github.com/angular/angular/issues/31712
     // As a work around, we ensuring Clarity access APIs outside of Zone (and use native implementation instead)
+    internal.log(Code.AngularZone, Severity.Info);
```
nairmanu: this will log AngularZone for every website. we need to only log when zone is available

SeenaAugusty: done

packages/clarity-js/src/core/api.ts
```diff
@@ -1,8 +1,14 @@
...
 export default function api(method: string): string {
     // Zone.js, a popular package for Angular, overrides native browser APIs which can lead to inconsistent state for single page applications.
     // Example issue: https://github.com/angular/angular/issues/31712
     // As a work around, we ensuring Clarity access APIs outside of Zone (and use native implementation instead)
-    return window[Constant.Zone] && Constant.Symbol in window[Constant.Zone] ? window[Constant.Zone][Constant.Symbol](method) : method;
+    const isZone = window[Constant.Zone] && Constant.Symbol in window[Constant.Zone];
+    if (isZone) {
+        internal.log(Code.AngularZone, Severity.Info);
+    }    
+    return  isZone ? window[Constant.Zone][Constant.Symbol](method) : method;
```
nairmanu: why the duplicate isZone check? cant we return `window[Constant.Zone][Constant.Symbol](method)` inside the if and then return method outside to simplify this?

nairmanu (changes requested):

packages/clarity-js/src/core/api.ts
SeenaAugusty: done

ender336 (approved):

PR #853, opened 2025-06-12
Visualization improvements (click visibility, iframe/masking clearness)

ender336:


packages/clarity-visualize/src/layout.ts
```diff
@@ -561,12 +666,15 @@ export class LayoutHelper {
...
-        return this.getImageHiddenCss() +
+        let retVal = this.getImageHiddenCss() +
             this.getIframeUnavailableCss() +
-            `${Constant.IFrameTag}[${Constant.UnavailableSmall}] { ${iframeUnavailableSmallSvg} }` +
+            this.getBlobUnavailableCss() +
+            `${Constant.IFrameTag}[${Constant.UnavailableSmall}] { ${iframeUnavailableSvgSmall} }` +
             `*[${Constant.Suspend}] { filter: grayscale(100%); }` +
             `body { font-size: initial; }
             ${this.getMobileCustomStyle()}`;
+        console.log('custom styles will be ', retVal);
```
malaref: Is this `console.log` intentional?

packages/clarity-visualize/src/layout.ts
```diff
@@ -579,16 +687,25 @@ export class LayoutHelper {
...
         if (this.vNext) {
-            return `${Constant.IFrameTag}[${Constant.Unavailable}] { ${iframeUnavailableSvg} }`;
+            return `${Constant.IFrameTag}[${Constant.UnavailableSmall}] { ${iframeUnavailableSvgSmall} }` +
+                    `${Constant.IFrameTag}[${Constant.Unavailable}] { ${iframeUnavailableSvg[this.locale]} }`;
         } else {
             return `${Constant.IFrameTag}[${Constant.Unavailable}] { background: url(${Asset.Unavailable}) no-repeat center center, url('${Asset.Cross}'); }`;
         }
     }
 
+    private getBlobUnavailableCss = (): string => {
```
malaref, line 696: This doesn't need to be a function. Probably a member initialized in the constructor (like `this.locale`) is better.

packages/clarity-visualize/src/layout.ts
ender336: no, good catch thank you

packages/clarity-visualize/src/layout.ts
ender336: Hmm why do you like it better initialized in the constructor? I could put the various css strings into members and initialize them, but I don't think it's any more efficient really - this only gets called once when the HEAD is found for rendering.

Moving to an internal member saves a function call later, but keeps the strings in memory longer so it feels like a wash to me.

malaref: I wasn't aware that function only gets called once. To me, it'd be better for readability (I try to keep all readonly constants marked as such), but I don't have a strong opinion on this.

PR #856, opened 2025-06-17
Changing AngularZone websites to send as metrics

SeenaAugusty:


packages/clarity-js/src/data/metadata.ts
```diff
@@ -49,10 +49,17 @@ export function start(): void {
...
 
-  // Capture additional metadata as metrics
-  metric.max(Metric.ClientTimestamp, s.ts);
-  metric.max(Metric.Playback, BooleanFlag.False);
-  metric.max(Metric.Electron, electron);
+    // Capture additional metadata as metrics
+    metric.max(Metric.ClientTimestamp, s.ts);
+    metric.max(Metric.Playback, BooleanFlag.False);
+    metric.max(Metric.Electron, electron);
+
```
nairmanu, line 57: why are we adding changes these lines? lets only include what is required

packages/clarity-js/types/data.d.ts
```diff
@@ -346,6 +347,8 @@ export const enum Constant {
     SHA256 = "SHA-256",
     Electron = "Electron",
     Caret = "^",
+    Zone = "Zone",
+    Symbol = "__symbol__",
```
nairmanu: why this change? isnt Symbol a constant and not a dimension?

packages/clarity-js/src/data/metadata.ts
```diff
@@ -49,10 +49,17 @@ export function start(): void {
...
+    const zone = (window as any)?.[Constant.Zone];
+    const isZone = zone && Constant.Symbol in zone;
```
nairmanu: import { Constant as CoreConstant } from "@clarity-types/core";

nairmanu (approved):

PR #858, opened 2025-06-18
Mn/rbiome

nairmanu:


packages/clarity-js/types/interaction.d.ts
```diff
@@ -15,7 +15,10 @@ export const enum Setting {
     LookAhead = 500, // 500ms
     InputLookAhead = 1000, // 1s
     Distance = 20, // 20 pixels
-    Interval = 25, // 25 milliseconds
+    ScrollInterval = 50, // 25 milliseconds
+    PointerInterval = 25, // 25 milliseconds
+    Throttle = 25, // 25 milliseconds
+    // biome-ignore lint/style/useLiteralEnumMembers: Time.Second is a const enum, it is compiled to a number
```
ender336: hmm this seems like it might be polluted from biome commit

nairmanu: yes, i choose to ignore and let it be because it was just a comment  :)

PR #884, opened 2025-06-30
Add defensive checks for nullable objects

AbdelrhmanMagdy:


packages/clarity-js/src/interaction/scroll.ts
```diff
@@ -33,6 +33,10 @@ function recompute(event: UIEvent = null): void {
     let de = document.documentElement;
     let element = event ? target(event) : de;
```
nairmanu, line 34: minor request: can u add one more fallback to document.body when document.documentElement is also not available?

nairmanu (approved):

PR #890, opened 2025-07-07
Refactor queue calls to include transmit parameter for upload compute()

nairmanu:


packages/clarity-js/src/interaction/timeline.ts
```diff
@@ -45,7 +45,7 @@ export function track(time: number,
 export function compute(): void {
     const temp = [];
     updates = [];
-    let max = envelope.data.start + envelope.data.duration;
+    let max = (envelope?.data?.start || 0) + (envelope?.data?.duration || 0);
```
ender336: I think it would be better to bail if envelop.data is null - we don't need to upload timeline in that case (it's gonna be garbage data)

ender336 (approved):

PR #894, opened 2025-07-10
Add defensive null checks across core modules

AbdelrhmanMagdy:


packages/clarity-js/src/performance/observer.ts
```diff
@@ -1,5 +1,5 @@
 import { Code, Constant, Dimension, Metric, Severity, PerformanceEventTiming } from "@clarity-types/data";
-import { FunctionNames } from "@clarity-types/performance";
+import { FunctionNames, NavigatorConnection } from "@clarity-types/performance";
```
malaref: This new import is never used.

packages/clarity-js/src/performance/observer.ts
```diff
@@ -15,7 +15,7 @@ const types: string[] = [Constant.Navigation, Constant.Resource, Constant.LongTa
 
 export function start(): void {
     // Capture connection properties, if available
-    if (navigator && "connection" in navigator) {
+    if (navigator && "connection" in navigator && navigator.connection) {
```
malaref: Isn't this check redundant?

packages/clarity-js/src/data/metadata.ts
```diff
@@ -21,7 +21,7 @@ let defaultStatus: ConsentState = {ad_Storage: Constant.Denied, analytics_Storag
 export function start(): void {
   rootDomain = null;
   const ua = navigator && "userAgent" in navigator ? navigator.userAgent : Constant.Empty;
-  const timezone = Intl?.DateTimeFormat()?.resolvedOptions()?.timeZone ?? '';
+  const timezone = (typeof Intl !== 'undefined' && Intl?.DateTimeFormat()?.resolvedOptions()?.timeZone) ?? '';
```
malaref, line 24: Isn't this check redundant?

packages/clarity-js/src/data/encode.ts
```diff
@@ -63,10 +63,12 @@ export default function(event: Event): void {
             queue(tokens);
             break;
         case Event.Upload:
-            tokens.push(track.sequence);
-            tokens.push(track.attempts);
-            tokens.push(track.status);
-            queue(tokens, false);
+            if (track) {
```
nairmanu: when will track be null? I only see from one place we call encode(Event.Upload)

track = { sequence, attempts: transitData.attempts, status: xhr.status };
            // Send back an event only if we were not successful in our first attempt
            if (transitData.attempts > 1) { encode(Event.Upload); }

based on that, track will never be null. Just making sure that we are cognizant of adding value when we add checks and not bloat codebase.

packages/clarity-js/src/interaction/click.ts
```diff
@@ -24,7 +24,7 @@ export function observe(root: Node): void {
 function handler(event: Event, root: Node, evt: MouseEvent): void {
     handler.dn = FunctionNames.ClickHandler;
     let frame = iframe(root);
-    let d = frame ? frame.contentDocument.documentElement : document.documentElement;
+    let d = frame && frame.contentDocument ? frame.contentDocument.documentElement : document.documentElement;
```
nairmanu, line 27: frame.contentDocument shouldnt be null here. iframe(root) returns the frame based on root -> the root is the contentDocument in this case. frame should be null if frame.contentDocument is null.

packages/clarity-js/src/interaction/pointer.ts
```diff
@@ -35,7 +35,7 @@ export function observe(root: Node): void {
 function mouse(event: Event, root: Node, evt: MouseEvent): void {
     mouse.dn = FunctionNames.PointerMouse;
     let frame = iframe(root);
-    let d = frame ? frame.contentDocument.documentElement : document.documentElement;
+    let d = frame && frame.contentDocument ? frame.contentDocument.documentElement : document.documentElement;
```
nairmanu, line 38: same here
frame.contentDocument shouldnt be null here. iframe(root) returns the frame based on root -> the root is the contentDocument in this case. frame should be null if frame.contentDocument is null.

packages/clarity-js/src/interaction/pointer.ts
```diff
@@ -52,7 +52,7 @@ function mouse(event: Event, root: Node, evt: MouseEvent): void {
 function touch(event: Event, root: Node, evt: TouchEvent): void {
     touch.dn = FunctionNames.PointerTouch;
     let frame = iframe(root);
-    let d = frame ? frame.contentDocument.documentElement : document.documentElement;
+    let d = frame && frame.contentDocument ? frame.contentDocument.documentElement : document.documentElement;
```
nairmanu, line 55: here too 
frame.contentDocument shouldnt be null here. iframe(root) returns the frame based on root -> the root is the contentDocument in this case. frame should be null if frame.contentDocument is null.

packages/clarity-js/src/data/metadata.ts
AbdelrhmanMagdy: I came across logs indicating that Intl isn't defined at all on a specific device (Nintendo Switch). Since optional chaining doesn't prevent a ReferenceError when the variable itself is undefined, I added the typeof Intl !== 'undefined' check to avoid runtime crashes in such cases.

packages/clarity-js/src/performance/observer.ts
AbdelrhmanMagdy: Some browser versions implement the navigator.connection interface, but the property itself is undefined. So attempting to access navigator.connection directly in those cases results in a runtime error.

AbdelrhmanMagdy: Replacing with navigator?.connection can cause a TypeScript error because connection is not officially included in the Navigator type definitions.

packages/clarity-js/src/data/encode.ts
AbdelrhmanMagdy: reverting this change for now.

PR #896, opened 2025-07-16
Add context menu event handling (0.8.20)

nairmanu:
This pull request introduces a new feature for handling `contextmenu` events in the Clarity library, updates dependencies and versions across packages, and adds tests to ensure the new functionality works as intended. Additionally, it removes the unused `.travis.yml` file.


Niyibitanga (approved):

packages/clarity-js/src/interaction/click.ts
```diff
@@ -19,6 +19,7 @@ export function start(): void {
 
 export function observe(root: Node): void {
     bind(root, "click", handler.bind(this, Event.Click, root), true);
+    bind(root, "contextmenu", handler.bind(this, Event.ContextMenu, root), true);
```
toby-walker, line 22: This is really more like a pointer/mouse event. Should it go there? Are event in click (more or less) a superset of the pointer event properties?

toby-walker, line 22: Please don't reuse an existing event ahere.

toby-walker (changes requested):
:clock1:

packages/clarity-js/src/interaction/encode.ts
```diff
@@ -53,6 +53,7 @@ export default async function (type: Event, ts: number = null): Promise<void> {
             pointer.reset();
             break;
         case Event.Click:
+        case Event.ContextMenu:
```
toby-walker, line 56: I don't see this used. Did I miss it?

packages/clarity-js/src/interaction/click.ts
nairmanu: Click event in a subset of pointer event properties. the properties we would read from a contextmenu are very similar to what we capture in click, hence the placement.

packages/clarity-decode/test/decode.test.js
```diff
@@ -0,0 +1,36 @@
+const { assert } = require('chai');
```
malaref, line 1: Why are tests using CommonJS `require` syntax instead of ECMAScript `import`?

packages/clarity-decode/test/decode.test.js
malaref, line 1: Why is this `.js` not `.ts` like the rest of tests in `clarity-js`?

yarn.lock
```diff
@@ -1682,6 +1687,19 @@ chai@^4.2.0:
     pathval "^1.1.1"
     type-detect "^4.0.5"
 
+chai@^4.3.0:
```
malaref, line 1672: This (and other testing libraries) are duplicated due to the mismatch between `clarity-js` and `clarity-decode` here.

malaref: I think `clarity-decode`'s testing share its setup with `clarity-js`. Maybe it can have its own PR.

packages/clarity-decode/test/decode.test.js
nairmanu: because our output is commonjs and im testing the final build, not the typescript

yarn.lock
nairmanu: i changed to ensure we use same versions across. now they are aligned

packages/clarity-decode/test/decode.test.js
nairmanu: because im testing the final build output and not the typescript code for decode

malaref (approved):

PR #914, opened 2025-08-08
Add support for dynamic loading of new modules

nairmanu:
This pull request introduces dynamic module loading capabilities to the codebase, allowing scripts to be loaded and managed at runtime. It adds a new `dynamic` module, exposes related functions for external use, and integrates dynamic event handling into the upload response flow.

### DIFF
--- a/packages/clarity-js/src/core/dynamic.ts
+++ b/packages/clarity-js/src/core/dynamic.ts
@@ -1,0 +1,42 @@
+let stopCallbacks: (() => void)[] = [];
+let active = false;
+
+export function start(): void {
+  active = true;
+}
+
+export function stop(): void {
+  stopCallbacks.reverse().forEach((callback) => {
+    try {
+      callback();
+    } catch (error) {
+      // Do nothing
+    }
+  });
+  stopCallbacks = [];
+  active = false;
+}
+
+export function register(stopCallback: () => void): void {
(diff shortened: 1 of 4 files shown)


packages/clarity-js/src/core/dynamic.ts
```diff
@@ -0,0 +1,42 @@
...
+}
+
+export function register(stopCallback: () => void): void {
+  if (active && typeof stopCallback === "function") {
+    stopCallbacks.push(stopCallback);
+  }
+}
+
+export function dynamicEvent(url: string): void {
+  if (!active) return;
```
Niyibitanga: Is this check necessary, if load does the same check?

malaref: I agree and maybe the two functions can be merged as one.

malaref: Maybe also change the name to `dynamicScript` or something like that, as this is not a typical event like `InputEvent`

malaref (approved):

packages/clarity-js/src/core/dynamic.ts
nairmanu: Good point! removed the redundant check. i kept the name as dynamicEvent to be consistent with the signalEvent which is also a response from server

Niyibitanga (approved):

PR #935, opened 2025-09-19
Optimize baseline naming for bundle size

nairmanu:


malaref:
Why time doesn't follow the same convention as the others? `tMove` instead of `moveT`?

toby-walker:
Old. Closing.

PR #937, opened 2025-09-24
Web Components Support (0.8.31)

malaref:
This pull request introduces support for tracking and replaying custom web components (`customElements`), alongside a version bump to `0.8.31`.

**Custom Web Components Tracking and Replay**
* Added `custom.ts` in `clarity-js` to hook into `window.customElements.define`, detect custom element registrations, and encode them as a new event type (`CustomElement`). It also checks existing nodes for any prior registrations of elements.
* Updated layout encoding/decoding logic in `clarity-js` and `clarity-decode` to handle the new `CustomElement` event.
* Added custom element replay support in `clarity-visualize`, allowing visualizer to define custom elements during session playback.

**TypeScript Configuration**
* Updated TypeScript target from `es5` to `es6` for `clarity-visualize` to support custom elements syntax.

**Version Updates:**
* Bumped version numbers from `0.8.30` to `0.8.31` in all relevant files.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -2,6 +2,6 @@
   "packages": [
     "packages/*"
   ],
-  "version": "0.8.30",
+  "version": "0.8.31",
   "npmClient": "yarn"
 }
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "clarity",
   "private": true,
-  "version": "0.8.30",
+  "version": "0.8.31",
   "repository": "https://github.com/microsoft/clarity.git",
   "author": "Sarvesh Nagpal <sarveshn@microsoft.com>",
   "license": "MIT",
--- a/packages/clarity-decode/package.json
+++ b/packages/clarity-decode/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clarity-decode",
-  "version": "0.8.30",
+  "version": "0.8.31",
   "description": "An analytics library that uses web page interactions to generate aggregated insights",
   "author": "Microsoft Corp.",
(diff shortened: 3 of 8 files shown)


packages/clarity-js/src/layout/encode.ts
```diff
@@ -120,6 +121,10 @@ export default async function (type: Event, timer: Timer = null, ts: number = nu
                 queue(tokenize(tokens), !config.lean);
             }
             break;
+        case Event.CustomElement:
+            tokens.push(custom.element);
+            queue(tokens);
```
nairmanu: shouldnt there be an update to queue() to use this event as playback event and not analytic event?

malaref: Good point, done!

packages/clarity-js/src/layout/custom.ts
```diff
@@ -0,0 +1,42 @@
...
+import * as core from "@src/core";
+
+export let element: string | undefined;
+
+const definedElements = new Set<string>();
+function register(tag: string) {
+    if (!definedElements.has(tag)) {
+        definedElements.add(tag);
+        element = tag;
+        encode(Event.CustomElement);
```
nairmanu: any reason not to use task.schedule here but execute encode immediately? wouldn't it be better to yield first?

malaref: Changed it to use `schedule`

ender336 (approved):

nairmanu (approved):

nairmanu (approved):

PR #951, opened 2025-10-09
Upgrade devtools manifest to V3

alj032:
This upgrades the devtools manifest to version 3, allowing it to run on modern browsers

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -2,6 +2,6 @@
   "packages": [
     "packages/*"
   ],
-  "version": "0.8.36",
+  "version": "0.8.38",
   "npmClient": "yarn"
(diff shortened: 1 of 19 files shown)


alj032:

alj032:
To be completely transparent here,  I am not very familiar with typescript or service workers. This did appear to work correctly though on every site I tested and think this could possibly be a valuable starting place for someone more familiar with these items. I am also very new to using github and submitting pull requests to a open source project, so my apologies if something is not quite right. Looking forward to feedback.

toby-walker:
Handle in this PR instead: #1002 .Thanks for contributing.

PR #955, opened 2025-10-17
Update Gcm to Check for Missed Consent

Niyibitanga:
Sometimes the consent state will be loaded before tags and it will be missed. Updating to check for consent updates prior to tags being loaded.

### DIFF
--- a/packages/clarity-js/src/data/consent.ts
+++ b/packages/clarity-js/src/data/consent.ts
@@ -62,14 +62,14 @@ export function trackConsentv2(consent: ConsentData): void {
     encode(Event.Consent);
 }
 
-// Compute function is called every upload, but we only want to send consent data once.
+// Compute function is called every upload, but we only want to send consent data once or if consent is updated.
 export function compute(): void {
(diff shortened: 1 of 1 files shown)


packages/clarity-js/src/data/consent.ts
```diff
@@ -66,5 +67,10 @@ export function compute(): void {
     if (updateConsent) {
         encode(Event.Consent);
         updateConsent = false;
+
+        const ics = window.google_tag_data?.ics;
+        if(ics?.usedUpdate){
```
nairmanu: fix styling issues

packages/clarity-js/src/data/consent.ts
```diff
@@ -66,5 +67,10 @@ export function compute(): void {
...
+            processConsent();
```
nairmanu, line 73: shouldnt u do this before encode and updateConsent is set to false?

nairmanu (approved):

nairmanu (approved):

PR #966, opened 2025-10-28
Enhance Consent Processing

Niyibitanga:
Update Consent Handling to only process changes


packages/clarity-js/src/data/metadata.ts
nairmanu: pls add a check here to make sure we dont reset cookie and restart clarity if track is already false

nairmanu (changes requested):

nairmanu (approved):

PR #986, opened 2025-12-02
Maintain Source at Restart

Niyibitanga:
Update to maintain consent source at restart

### DIFF
--- a/packages/clarity-js/src/data/metadata.ts
+++ b/packages/clarity-js/src/data/metadata.ts
@@ -189,7 +189,7 @@ export function consentv2(consentState: ConsentState = defaultStatus, source: nu
 
 function getConsentData(consentState: ConsentState): ConsentData {
   let consent: ConsentData = {
-    source: consentState.source,
+    source: consentState.source ?? ConsentSource.Implicit,
     ad_Storage: consentState.ad_Storage === Constant.Granted ? BooleanFlag.True : BooleanFlag.False,
     analytics_Storage: consentState.analytics_Storage === Constant.Granted ? BooleanFlag.True : BooleanFlag.False,
   };


packages/clarity-js/src/data/metadata.ts
```diff
@@ -150,6 +151,7 @@ export function consent(status = true): void {
 
 export function consentv2(consentState: ConsentState = defaultStatus, source: number = ConsentSource.API): void {
   const updatedStatus = {
+    source: source,
```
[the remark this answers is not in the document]
Niyibitanga: next iteration

packages/clarity-js/src/data/metadata.ts
```diff
@@ -16,7 +16,7 @@ export let callbacks: MetadataCallbackOptions[] = [];
 export let electron = BooleanFlag.False;
 let rootDomain = null;
 let consentStatus: ConsentState = null;
-let defaultStatus: ConsentState = { ad_Storage: Constant.Denied, analytics_Storage: Constant.Denied };
+let defaultStatus: ConsentState = { source: ConsentSource.API, ad_Storage: Constant.Denied, analytics_Storage: Constant.Denied };
```
nairmanu, line 19: why default to API?

Niyibitanga: the object is only used if there is nothing passed in the API call

packages/clarity-js/src/data/metadata.ts
```diff
@@ -16,7 +16,7 @@ export let callbacks: MetadataCallbackOptions[] = [];
 export let electron = BooleanFlag.False;
 let rootDomain = null;
 let consentStatus: ConsentState = null;
```
toby-walker, line 18: >consentStatus [](http://example.com/codeflow?start=4&length=13)

Could we separate APII source into V1 and V2? It''s good know which is which.

packages/clarity-js/src/data/metadata.ts
```diff
@@ -185,9 +187,9 @@ export function consentv2(consentState: ConsentState = defaultStatus, source: nu
   trackConsent.consent();
 }
 
-function getConsentData(consentState: ConsentState, source: ConsentSource): ConsentData {
+function getConsentData(consentState: ConsentState): ConsentData {
```
toby-walker, line 190: >getConsentData [](http://example.com/codeflow?start=9&length=14)

When would the ?? happen?

toby-walker:
>     consentStatus &&

Is it possible send an event in this case unless source == last source. No enforcement required. We want to track consent from all source event if it is the same. Critical for debugging consent  issues. Can be another PR later.

It would be good to separate the logic for tracking from enforcement (doing something with it).

---
Refers to: packages/clarity-js/src/data/metadata.ts:160 in 5fa673e. [](commit_id = 5fa673efcdb564d9bbddb35ea4179df09153de83, deletion_comment = False)

packages/clarity-js/src/data/metadata.ts
Niyibitanga: it wouldn't happen, all functions that call it pass the value. But just being extra safe in case later update/changes don't pass it. Theoretically should never happen

malaref: I also don't see a path for `??` to matter, so I'd remove it to disallow/catch later changes not passing it

malaref (approved):

packages/clarity-js/src/data/metadata.ts
Niyibitanga: next iteration

nairmanu (approved):

PR #992, opened 2025-12-09
Update Consent Storage

Niyibitanga:
Update to store consent type as the values used for processing instead of object sent by site.

### DIFF
--- a/packages/clarity-js/src/data/metadata.ts
+++ b/packages/clarity-js/src/data/metadata.ts
@@ -15,7 +15,7 @@ import { set } from "@src/data/variable";
 export let data: Metadata = null;
 export let callbacks: MetadataCallbackOptions[] = [];
 export let electron = BooleanFlag.False;
-let consentStatus: ConsentState = null;
+let consentStatus: ConsentData = null;
 let cookiesLogged = false;
 let defaultStatus: ConsentState = { source: ConsentSource.Default, ad_Storage: Constant.Denied, analytics_Storage: Constant.Denied };
 
@@ -84,15 +84,14 @@ export function start(): void {
   if (consentStatus === null) {
     consentStatus = {
       source: u.consent ? ConsentSource.Cookie : ConsentSource.Implicit,
-      ad_Storage: config.track ? Constant.Granted : Constant.Denied,
-      analytics_Storage: config.track ? Constant.Granted : Constant.Denied,
+      ad_Storage: config.track ? BooleanFlag.True : BooleanFlag.False,
+      analytics_Storage: config.track ? BooleanFlag.True : BooleanFlag.False,
     };
   }
 
   logCookies();
 
-  const consent = getConsentData(consentStatus);
-  trackConsent.config(consent);
+  trackConsent.config(consentStatus);
   // Track ids using a cookie if configuration allows it
   track(u);
 }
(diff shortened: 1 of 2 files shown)


packages/clarity-js/src/data/metadata.ts
```diff
@@ -15,7 +15,7 @@ export let data: Metadata = null;
 export let callbacks: MetadataCallbackOptions[] = [];
 export let electron = BooleanFlag.False;
 let rootDomain = null;
-let consentStatus: ConsentState = null;
+let consentStatus: ConsentData = null;
 let defaultStatus: ConsentState = { source: ConsentSource.API, ad_Storage: Constant.Denied, analytics_Storage: Constant.Denied };
```
toby-walker: Should this not be ConsentData type?  Maybe also consider giving this a unique source as well (the more the better).

packages/clarity-js/src/data/metadata.ts
```diff
@@ -147,15 +147,18 @@ export function consent(status = true): void {
...
   consentv2({ ad_Storage: Constant.Granted, analytics_Storage: Constant.Granted });
-  trackConsent.consent();
 }
 
 export function consentv2(consentState: ConsentState = defaultStatus, source: number = ConsentSource.API): void {
-  const updatedStatus = {
+
+  const currentState = getConsentState();
+  const consentUpdate = {
     source: consentState.source ?? source,
```
toby-walker: Wouldn't this alway mean we keep the last source? Should it not be flipped?

packages/clarity-js/src/data/metadata.ts
```diff
@@ -147,15 +147,18 @@ export function consent(status = true): void {
   }
 
   consentv2({ ad_Storage: Constant.Granted, analytics_Storage: Constant.Granted });
-  trackConsent.consent();
 }
 
 export function consentv2(consentState: ConsentState = defaultStatus, source: number = ConsentSource.API): void {
```
toby-walker: In this or later PR change to APIV2 so we know more precisely the source.

packages/clarity-js/src/data/metadata.ts
```diff
@@ -198,6 +200,14 @@ function getConsentData(consentState: ConsentState): ConsentData {
   return consent;
 }
 
+function getConsentState(): ConsentState {
+  return {
+    source: consentStatus.source ?? ConsentSource.Implicit,
```
toby-walker: When can source be null/undefined?

packages/clarity-js/src/data/metadata.ts
Niyibitanga: ConsentData uses boolean flags, the default status uses strings "granted" | "denied" sent to the API

packages/clarity-js/src/data/metadata.ts
```diff
@@ -158,14 +157,14 @@ export function consent(status = true): void {
...
   if (!core.active()) { return; }
 
-  const updatedStatus = {
-    source: consentState.source ?? source,
-    ad_Storage: normalizeConsent(consentState.ad_Storage, consentStatus?.ad_Storage),
-    analytics_Storage: normalizeConsent(consentState.analytics_Storage, consentStatus?.analytics_Storage),
+  const updatedStatus: ConsentData = {
+    source: consentInput.source ?? source,
+    ad_Storage: normalizeConsent(consentInput.ad_Storage, consentStatus?.ad_Storage),
+    analytics_Storage: normalizeConsent(consentInput.analytics_Storage, consentStatus?.analytics_Storage),
```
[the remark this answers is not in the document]
Niyibitanga: getConsentState() is only called after start() has initialized consentStatus — all call sites are gated by core.active() or data being set. This is consistent with getConsentData() which follows the same pattern without a null check.

PR #993, opened 2025-12-09
Send Updates When No Changes To Consent

Niyibitanga:
Updating API to send consent updates even when there are no changes to permissions.

### DIFF
--- a/packages/clarity-js/src/data/consent.ts
+++ b/packages/clarity-js/src/data/consent.ts
@@ -30,11 +30,11 @@ function processConsent(): void {
 
     const analytics_storage = ics.getConsentState(Constant.AnalyticsStorage);
     const ad_storage = ics.getConsentState(Constant.AdStorage);
-    const consentState = getConsentState({ ad_Storage: ad_storage, analytics_Storage: analytics_storage });
+    const consentState = getGcmConsentState({ ad_Storage: ad_storage, analytics_Storage: analytics_storage });
     consentv2(consentState);
 }
 
-function getConsentState(googleConsent: GCMConsentState): ConsentState {
+function getGcmConsentState(googleConsent: GCMConsentState): ConsentState {
     const consentState: ConsentState = {
         source: ConsentSource.GCM,
         ad_Storage: googleConsent.ad_Storage === GCMConsent.Granted ? Constant.Granted : Constant.Denied,


packages/clarity-js/src/data/metadata.ts
```diff
@@ -162,6 +162,9 @@ export function consentv2(consentState: ConsentState = defaultStatus, source: nu
     updatedStatus.ad_Storage === consentStatus.ad_Storage &&
     updatedStatus.analytics_Storage === consentStatus.analytics_Storage
   ) {
+    consentStatus.source = updatedStatus.source;
+    trackConsent.trackConsentv2(getConsentData(consentStatus));
```
toby-walker, line 164: Could not send event when source == previous source.

Niyibitanga: in what scenario?

packages/clarity-js/src/data/consent.ts
```diff
@@ -31,11 +31,12 @@ function processConsent(): void {
     const analytics_storage = ics.getConsentState(Constant.AnalyticsStorage);
     const ad_storage = ics.getConsentState(Constant.AdStorage);
     const consentState = getConsentState({ ad_Storage: ad_storage, analytics_Storage: analytics_storage });
-    consentv2(consentState, ConsentSource.GCM);
+    consentv2(consentState);
 }
 
 function getConsentState(googleConsent: GCMConsentState): ConsentState {
     const consentState: ConsentState = {
+        source: ConsentSource.GCM,
```
toby-walker, line 39: Minor; Name is confusing and too general. Should be getGcmConsentState

packages/clarity-js/src/data/metadata.ts
```diff
@@ -160,6 +160,9 @@ export function consentv2(consentState: ConsentState = defaultStatus, source: nu
...
+    trackConsent.consent();
```
nairmanu, line 165: Why are we tracking old consent as well? cant we just do the encode?

packages/clarity-js/src/data/consent.ts
nairmanu, line 58: do we need this anymore? can we do away with it and just use consentv2?

Niyibitanga: There is BE logic utilizing it, maybe we can include it in one of my other clean/refactor prs

packages/clarity-js/src/data/metadata.ts
Niyibitanga: calling encode would have the old source value, which defeats the purpose of this change

PR #995, opened 2025-12-10
Extract cookie handling to separate module (0.8.44)

nairmanu:
This pull request refactors cookie handling in the Clarity JS codebase by moving all cookie-related logic from `metadata.ts` to a new dedicated module, `cookie.ts`. It also introduces utility functions in `util.ts` and updates the initialization sequence to include the new cookie module. These changes improve code organization, reusability, and maintainability.

**Refactoring and modularization:**
* Extracted all cookie-related logic (`getCookie`, `setCookie`, encoding/decoding, and domain handling) from `metadata.ts` into a new `cookie.ts` module, centralizing cookie management. 

**Integration and initialization:**
* Registered the new `cookie` module in the main data module list and ensured it is started with other modules. 

**Consent and session management improvements:**
* Updated the `clear` and `consentv2` functions to use the new cookie module for clearing cookies, and enhanced the `clear` function to optionally clear all relevant cookies.

### DIFF
--- a/packages/clarity-js/src/data/metadata.ts
+++ b/packages/clarity-js/src/data/metadata.ts
@@ -6,7 +6,7 @@ import config from "@src/core/config";
 import hash from "@src/core/hash";
 import * as scrub from "@src/core/scrub";
 import * as trackConsent from "@src/data/consent";
-import { getCookie, setCookie } from "@src/data/cookie";
+import { COOKIE_SEP, getCookie, setCookie } from "@src/data/cookie";
 import * as dimension from "@src/data/dimension";
 import * as metric from "@src/data/metric";
 import { supported } from "@src/data/util";
@@ -229,7 +229,7 @@ export function save(): void {
   let ts = Math.round(Date.now());
   let upload = config.upload && typeof config.upload === Constant.String ? (config.upload as string).replace(Constant.HTTPS, Constant.Empty) : Constant.Empty;
   let upgrade = config.lean ? BooleanFlag.False : BooleanFlag.True;
-  setCookie(Constant.SessionKey, [data.sessionId, ts, data.pageNum, upgrade, upload].join(Constant.Caret), Setting.SessionExpire);
+  setCookie(Constant.SessionKey, [data.sessionId, ts, data.pageNum, upgrade, upload].join(COOKIE_SEP), Setting.SessionExpire);
 }
 
 function processCallback(upgrade: BooleanFlag, consentUpdate: boolean = false): void {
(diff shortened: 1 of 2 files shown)


packages/clarity-js/src/data/metadata.ts
```diff
@@ -202,9 +199,14 @@ function normalizeConsent(value: unknown, fallback: string = Constant.Denied): s
   return typeof value === 'string' ? value.toLowerCase() : fallback;
 }
 
-export function clear(): void {
+export function clear(all: boolean = false): void {
   // Clear any stored information in the cookie that tracks session information so we can restart fresh the next time
-  setCookie(Constant.SessionKey, Constant.Empty, 0);
+  setCookie(Constant.SessionKey, Constant.Empty, -Number.MAX_VALUE);
```
malaref: Why the change to `-Number.MAX_VALUE` instead of `0`? This may change the behavior of `setCookie` as `cookie` will be `${key}=${encodedValue};expires=Invalid Date;path=/`;

packages/clarity-js/src/data/metadata.ts
```diff
@@ -202,9 +199,14 @@ function normalizeConsent(value: unknown, fallback: string = Constant.Denied): s
   return typeof value === 'string' ? value.toLowerCase() : fallback;
 }
 
-export function clear(): void {
+export function clear(all: boolean = false): void {
```
ender336, line 202: I feel like this is a good opportunity to provide some more information in the naming so at a glance the behavior is clear. Maybe name the function clearCookies with two params (one for session which defaults to true and another for user which defaults to false)

ender336 (approved):

packages/clarity-js/src/data/metadata.ts
nairmanu: the outcome is the same: the cookie won’t persist. The difference is that 0 is a conventional way to clear a cookie, while -Number.MAX_VALUE is more of a hacky way to guarantee the expiry date is way before now. Since it was Number.MAX_VALUE where it was moslty used (in consent), I kept it consistent with that.

packages/clarity-js/src/data/metadata.ts
nairmanu: Exported function names wont get minified, so a longer name does impact bundle size :)

packages/clarity-js/src/data/metadata.ts
malaref: My point is that this number won't make it in the first place. `expiry.toUTCString()` will return `"Invalid Date"`.

nairmanu: it wont even get there. expiry will be NaN when we subtract the large number essentially turning the cookie to be a session cookie with no expiry. Setting to 0 would mean expiry will be current date but this is dependent on the browser returning correct value for new Date().toUTCString(). If local date is incorrect, we will not be compliant

Niyibitanga (approved):

PR #996, opened 2025-12-10
ensuring we don't process any script tags that were maliciously uploaded

ender336:


nairmanu (approved):

packages/clarity-visualize/src/layout.ts
```diff
@@ -452,6 +452,13 @@ export class LayoutHelper {
                     this.setAttributes(iframeElement, node);
                     insert(node, parent, iframeElement, pivot);
                     break;
+                case "SCRIPT":
+                    {
```
malaref, line 456: Why is this scoped, unlike other `case`s?... I don't think it'll affect the `break;` statement, but it still looks awkward 😅

ender336: @copilot All blocks should actually be scoped so that defined variables don't interfere with each other unintentionally. I just didn't want to make the change for all other fields as we wanted to get this in separately

toby-walker, line 456: Make comment a bit clearer. Give an example of what this is defending against.

toby-walker: Just merge the copilot PR that does this everywhere before your PR.

PR #998, opened 2025-12-12
Update Cookie Storage

Niyibitanga:
Update cookie consent to store ad storage as well

### DIFF
--- a/packages/clarity-js/src/data/metadata.ts
+++ b/packages/clarity-js/src/data/metadata.ts
@@ -88,9 +88,9 @@ export function start(): void {
   // If consent status is not already set, initialize it based on project configuration. Otherwise, use the existing consent status.
   if (consentStatus === null) {
     consentStatus = {
-      source: ConsentSource.Implicit,
-      ad_Storage: config.track ? Constant.Granted : Constant.Denied,
-      analytics_Storage: config.track ? Constant.Granted : Constant.Denied,
+      source: u.consent || u.ad_consent ? ConsentSource.Cookie : ConsentSource.Implicit,
+      ad_Storage: u.ad_consent || (config.track && consentStatus.source != ConsentSource.Cookie) ? Constant.Granted : Constant.Denied,
+      analytics_Storage: u.consent || config.track ? Constant.Granted : Constant.Denied,
     };
   }
   const consent = getConsentData(consentStatus);
@@ -177,7 +177,7 @@ export function consentv2(consentState: ConsentState = defaultStatus, source: nu
 
   if (core.active() && consentData.analytics_Storage) {
     config.track = true;
-    track(user(), BooleanFlag.True);
+    track(user(), consentData.analytics_Storage, consentData.ad_Storage);
     save();
   }
 
@@ -252,9 +252,10 @@ function processCallback(upgrade: BooleanFlag, consentUpdate: boolean = false): 
   }
 }
 
-function track(u: User, consent: BooleanFlag = null): void {
+function track(u: User, consent: BooleanFlag = null, ad_consent: BooleanFlag = null): void {
   // If consent is not explicitly specified, infer it from the user object
   consent = consent === null ? u.consent : consent;
+  ad_consent = ad_consent === null ? u.ad_consent : ad_consent;
   // Convert time precision into days to reduce number of bytes we have to write in a cookie
   // E.g. Math.ceil(1628735962643 / (24*60*60*1000)) => 18852 (days) => ejo in base36 (13 bytes => 3 bytes)
   let end = Math.ceil((Date.now() + (Setting.Expire * Time.Day)) / Time.Day);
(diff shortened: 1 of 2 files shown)


packages/clarity-js/types/data.d.ts
```diff
@@ -423,6 +424,7 @@ export interface User {
     version: number;
     expiry: number;
     consent: BooleanFlag;
```
nairmanu: since ad_consent is separate, shouldnt this be changed to analyticsConsent.

packages/clarity-js/src/data/metadata.ts
```diff
@@ -310,7 +311,8 @@ function user(): User {
     if (parts.length > 2) { output.expiry = num(parts[2], 36); }
     // Check if we have explicit consent to track this user
     if (parts.length > 3 && num(parts[3]) === 1) { output.consent = BooleanFlag.True; }
-    if (parts.length > 4 && num(parts[1]) > 1) { output.dob = num(parts[4]); }
+    if (parts.length > 4 && num(parts[4]) === 1) { output.ad_consent = BooleanFlag.True; }
```
nairmanu: we should not do this.  new parts should be at the end or there wont be any backward compatibility

packages/clarity-js/src/data/metadata.ts
```diff
@@ -88,9 +88,9 @@ export function start(): void {
   // If consent status is not already set, initialize it based on project configuration. Otherwise, use the existing consent status.
   if (consentStatus === null) {
     consentStatus = {
-      source: ConsentSource.Implicit,
-      ad_Storage: config.track ? Constant.Granted : Constant.Denied,
-      analytics_Storage: config.track ? Constant.Granted : Constant.Denied,
+      source: u.consent || u.ad_consent ? ConsentSource.Cookie : ConsentSource.Implicit,
+      ad_Storage: u.ad_consent || (config.track && consentStatus.source != ConsentSource.Cookie) ? Constant.Granted : Constant.Denied,
```
nairmanu: why the additional check for source? we shouldnt be checking config.track with this approach. There should be a ad and analytics value check instead

packages/clarity-js/types/data.d.ts
```diff
@@ -386,7 +386,8 @@ export const enum XMLReadyState {
 export const enum ConsentSource {
     Implicit = 0,
     API = 1,
-    GCM = 2
```
toby-walker, line 389: Could we add APIV1 vs. APIV2 at the same time or in a separate PR (if you prefer that).

packages/clarity-js/src/data/metadata.ts
```diff
@@ -88,9 +88,9 @@ export function start(): void {
...
+      analytics_Storage: u.consent || config.track ? Constant.Granted : Constant.Denied,
```
toby-walker, line 93: >i [](http://example.com/codeflow?start=42&length=1)

What is u.consent storing?

toby-walker, line 93: >t [](http://example.com/codeflow?start=57&length=1)

Style nit: Put parenthesis around the ?: to make the precedence clear.

packages/clarity-js/src/data/metadata.ts
toby-walker, line 92: Can you explain this logic?

toby-walker:
> let consentStatus: ConsentState = null;

Style nit: This name is a bit confusing. Better would be something like "currentCurreent' consent or similar. Unify everything to state not status as well in naming things. 

---
Refers to: packages/clarity-js/src/data/metadata.ts:18 in 46911be. [](commit_id = 46911bedd3fc11a6eb9e31ace91b26aea1abe45b, deletion_comment = False)

packages/clarity-js/src/data/metadata.ts
```diff
@@ -177,7 +177,7 @@ export function consentv2(consentState: ConsentState = defaultStatus, source: nu
 
   if (core.active() && consentData.analytics_Storage) {
     config.track = true;
-    track(user(), BooleanFlag.True);
+    track(user(), consentData.analytics_Storage, consentData.ad_Storage);
```
toby-walker, line 180: Minor: Might be clean to just pass the cseontnData as a single argument not splitting it up here into separate arg (enables us to add more permissions later). And lot of bool parameters are easy to get wrong (order, etc.).

packages/clarity-js/src/data/metadata.ts
```diff
@@ -252,18 +252,19 @@ function processCallback(upgrade: BooleanFlag, consentUpdate: boolean = false):
   }
 }
 
-function track(u: User, consent: BooleanFlag = null): void {
```
toby-walker, line 255: Rename to analytics to make it unambiguous. Consent own its own is too ambiguous.

packages/clarity-js/src/data/metadata.ts
```diff
@@ -252,18 +252,19 @@ function processCallback(upgrade: BooleanFlag, consentUpdate: boolean = false):
...
   // E.g. Math.ceil(1628735962643 / (24*60*60*1000)) => 18852 (days) => ejo in base36 (13 bytes => 3 bytes)
   let end = Math.ceil((Date.now() + (Setting.Expire * Time.Day)) / Time.Day);
   // If DOB is not set in the user object, use the date set in the config as a DOB
   let dob = u.dob === 0 ? (config.dob === null ? 0 : config.dob) : u.dob;
 
   // To avoid cookie churn, write user id cookie only once every day
-  if (u.expiry === null || Math.abs(end - u.expiry) >= Setting.CookieInterval || u.consent !== consent || u.dob !== dob) {
-    let cookieParts = [data.userId, Setting.CookieVersion, end.toString(36), consent, dob];
+  if (u.expiry === null || Math.abs(end - u.expiry) >= Setting.CookieInterval || u.consent !== consent || u.ad_consent != ad_consent || u.dob !== dob) {
+    let cookieParts = [data.userId, Setting.CookieVersion, end.toString(36), consent, ad_consent, dob];
```
toby-walker, line 267: This is not backwards/forwards compatible. Better to append to end. Think about how this work in current client (or older) and newer ones in all combination (older cookie, new client) and (new cookies, old client).

PR #1005, opened 2025-12-19
ensuring we dont execute code in attributes

ender336:


nairmanu (approved):

packages/clarity-visualize/src/layout.ts
```diff
@@ -670,6 +670,23 @@ export class LayoutHelper {
...
+        if (name.startsWith('on')) {
+            return true;
+        }
+        
+        // Check for JavaScript protocols and dangerous patterns
+        const dangerous = [
+            /^\s*javascript:/i,
+            /^\s*data:text\/html/i,
+            /^\s*vbscript:/i
+        ];
```
malaref, line 684: nit: this list should be among other constants

ender336: I think it's only relevant here so keeping it for now

PR #1017, opened 2026-01-18
Capture source information for clicks

nairmanu:


packages/clarity-js/src/interaction/click.ts
```diff
@@ -75,6 +75,7 @@ function handler(event: Event, root: Node, evt: MouseEvent): void {
                 tag: getElementAttribute(t, "tagName").substring(0, Setting.ClickTag),
                 class: getElementAttribute(t, "className").substring(0, Setting.ClickClass),
                 id: getElementAttribute(t, "id").substring(0, Setting.ClickId),
+                source: evt.isTrusted ? ClickSource.Undefined : source()
```
Niyibitanga: would it be better to have this as Clicksource.Trusted? Undefined seems a bit misleading

Niyibitanga (approved):

packages/clarity-js/src/interaction/click.ts
nairmanu: source having value trusted doesnt make sense. its just not defined in this case when isTrust is true.

packages/clarity-js/src/interaction/click.ts
```diff
@@ -76,6 +79,7 @@ function handler(event: Event, root: Node, evt: MouseEvent): void {
                 tag: getElementAttribute(t, "tagName").substring(0, Setting.ClickTag),
                 class: getElementAttribute(t, "className").substring(0, Setting.ClickClass),
                 id: getElementAttribute(t, "id").substring(0, Setting.ClickId),
+                source: config.diagnostics ? (evt.isTrusted ? ClickSource.Undefined : source()) : ClickSource.Undefined
```
nairmanu: this can be simplified to 
`config.diagnostics && !evt.isTrusted  ? source() : ClickSource.Undefined `

Niyibitanga (approved):

PR #1025, opened 2026-01-24
Fix(interaction): handle null targets in scroll/click to prevent Type Error #852

### LINKED ISSUE #852: Uncaught TypeError: Cannot read properties of null (reading 'scrollLeft')
**Describe the bug**
We are capturing hundreds of instances of this error via our Sentry integration within our PHP/Symfony app. When our users are navigating around our app, this error is being thrown which causes our front-end to break and throw an exception.

**Repro Steps**
Occurs in multiple scenarios within the UI of our PHP/Symfony app, when scrolling up/down within different components. Once the error is thrown in the front-end it bubbles up to the user through Symfony.

**Screenshots**
Image

Image

**Project info**
- PHP/Symfony application 

**Browser info:**
 - Principally in Windows using Chrome/Edge browser

BryanBradfo:
### Description
This PR fixes a `TypeError` reported in issue #852 ("Cannot read properties of null (reading 'scrollLeft')"). 

This error occurs when `scrollLeft` or `scrollTop` is accessed on a `null` element, which can happen in certain edge cases (e.g., inside React/Next.js applications where elements might be detached or managing shadow DOM).

### Changes
I added optional chaining (`?.`) and a fallback (`|| 0`) to handle null targets safely in:
- `packages/clarity-js/src/interaction/scroll.ts`
- `packages/clarity-js/src/interaction/click.ts`
- `packages/clarity-js/src/interaction/pointer.ts`

### Verification
- Ran `npm run build` successfully.
- Verified that the changes introduce null-checks without altering the logic for valid elements.

Fixes #852

cc @AbdelrhmanMagdy (Since you looked into this previously, I thought you might want to review this fix).

### DIFF
--- a/packages/clarity-js/src/interaction/click.ts
+++ b/packages/clarity-js/src/interaction/click.ts
@@ -24,8 +24,8 @@ export function observe(root: Node): void {
 function handler(event: Event, root: Node, evt: MouseEvent): void {
     let frame = iframe(root);
     let d = frame && frame.contentDocument ? frame.contentDocument.documentElement : document.documentElement;
-    let x = "pageX" in evt ? Math.round(evt.pageX) : ("clientX" in evt ? Math.round(evt["clientX"] + d.scrollLeft) : null);
-    let y = "pageY" in evt ? Math.round(evt.pageY) : ("clientY" in evt ? Math.round(evt["clientY"] + d.scrollTop) : null);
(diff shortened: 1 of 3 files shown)


BryanBradfo:

BryanBradfo:
Hi @malaref, @Niyibitanga and @nairmanu,

I'm bumping this PR as it addresses a TypeError (issue #852) affecting production environments in React/Next.js apps.

All checks have passed, the CLA is signed, and the fix uses safe optional chaining. Could you please take a look when you have a moment?

Thanks!

PR #1026, opened 2026-01-27
Enhance attribute handling by adding tag parameter for link attributes

nairmanu:
This pull request makes targeted improvements to attribute encoding logic. The main focus is on ensuring correct handling of the `href` attribute for link tags and improving privacy-related scrubbing.

**Attribute Encoding Improvements**

* Modified the attribute encoding logic in `clarity-js/src/insight/encode.ts` and `clarity-js/src/layout/encode.ts` so that when encoding the `href` attribute for link tags, the value is used directly without privacy scrubbing. This ensures that link URLs are preserved as expected for anchor elements.

**Privacy Scrubbing Logic**

* Simplified the privacy scrubbing for image and blob URLs in `scrub.ts` by using a conditional expression for `blob:` URLs, ensuring correct handling for privacy-sensitive attributes.

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -2,6 +2,6 @@
   "packages": [
     "packages/*"
   ],
-  "version": "0.8.51",
+  "version": "0.8.52",
   "npmClient": "yarn"
 }
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "clarity",
   "private": true,
-  "version": "0.8.51",
+  "version": "0.8.52",
   "repository": "https://github.com/microsoft/clarity.git",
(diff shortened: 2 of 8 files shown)


packages/clarity-js/src/core/scrub.ts
```diff
@@ -44,10 +44,7 @@ export function text(value: string, hint: string, privacy: Privacy, mangle: bool
                     case "href":
                     case "xlink:href":
                         if (privacy === Privacy.TextImage) {
-                            if (hint === 'src' && value?.startsWith('blob:')) {
```
rgaleev, line 47: is this condition `hint === 'src'` not needed anymore?

PR #1027, opened 2026-01-28
Update To Read Cookies Based On Client Consent

Niyibitanga:


packages/clarity-js/src/data/metadata.ts
```diff
@@ -16,6 +16,7 @@ export let data: Metadata = null;
 export let callbacks: MetadataCallbackOptions[] = [];
 export let electron = BooleanFlag.False;
 let consentStatus: ConsentState = null;
+let cookiesLogged = false;
```
toby-walker, line 19: Just confirming this is supposed be stateful and persist across restarts, etc. In other words it's not reset on module stop.() #Closed

packages/clarity-js/src/data/metadata.ts
```diff
@@ -93,6 +88,10 @@ export function start(): void {
       analytics_Storage: config.track ? Constant.Granted : Constant.Denied,
     };
   }
+
+  // Log cookies specified in configuration only if analytics_Storage is granted
+  if (consentStatus.analytics_Storage === Constant.Granted) { logCookies(); }
```
toby-walker: Some of these are ad related. Should we require both of these? Also might be a bit cleaner to move the check and log into a single function that first checks consent and then logs them. But this is ok as is. #Closed

packages/clarity-js/src/data/metadata.ts
```diff
@@ -112,8 +111,18 @@ function userAgentData(): void {
...
+  for (let key of config.cookies) {
+    let value = getCookie(key);
+    if (value) { set(key, value); }
+  }
+  cookiesLogged = true;
+}
+
 export function stop(): void {
   data = null;
+  cookiesLogged = false;
```
toby-walker, line 127: Should this be persistent and kept across restarts like consent state? #Closed

packages/clarity-js/src/data/metadata.ts
Niyibitanga: No, the cookies should be reread when a new session starts so that they are associated with the new session and it's data.

packages/clarity-js/src/data/metadata.ts
Niyibitanga: No, the cookies should be reread when a new session starts so that they are associated with the new session and it's data.

packages/clarity-js/src/data/metadata.ts
Niyibitanga: I agree, updated

PR #1031, opened 2026-01-30
Bump Version 0.8.53

Niyibitanga:


package.json
```diff
@@ -19,13 +19,13 @@
     "build:devtools": "yarn workspace clarity-devtools build",
     "test": "playwright test",
     "test:ui": "playwright test --ui",
-    "bump-version": "ts-node scripts/bump-version.ts"
```
nairmanu, line 22: why this change?

Niyibitanga: It fixes the issue with bump version command not working. It's been broken for a while. But we can deal with that later. I removed the changes.

nairmanu (approved):

PR #1033, opened 2026-02-03
Add Clarity Active Check for Consent Processing

Niyibitanga:


packages/clarity-js/src/data/metadata.ts
```diff
@@ -160,6 +163,9 @@ export function consent(status = true): void {
 }
 
 export function consentv2(consentState: ConsentState = defaultStatus, source: number = ConsentSource.APIv2): void {
+  // Guard against calling consent API when Clarity hasn't started (e.g., due to GPC)
+  if (!core.active()) { return; }
```
nairmanu, line 163: all we need is this. everything else is code bloat

PR #1042, opened 2026-02-18
[AI Visibility] AFD Integration

HusseinYoussef:
Pushing the UI template, bicep template and the compiled ARM template for AFD integration

image

### DIFF
--- a/devops/templates/ai-visibility/forward-afd-traffic/createUiDefinition.json
+++ b/devops/templates/ai-visibility/forward-afd-traffic/createUiDefinition.json
@@ -67,7 +67,6 @@
         "type": "Microsoft.Common.TextBox",
         "label": "Clarity Project ID",
         "toolTip": "Your Clarity project ID where bot detection data will be sent.",
-        "placeholder": "abc123xyz",
         "constraints": {
           "required": true,
           "validations": [


HusseinYoussef:

devops/templates/ai-visibility/forward-afd-traffic/azuredeploy.json
```diff
@@ -0,0 +1,300 @@
...
+              "name": "eventhub-connection",
+              "value": "[listKeys(format('{0}/authorizationRules/RootManageSharedAccessKey', resourceId('Microsoft.EventHub/namespaces', format('{0}-ehns', variables('namePrefix')))), '2023-01-01-preview').primaryConnectionString]"
+            }
+          ]
+        },
+        "template": {
+          "containers": [
+            {
+              "name": "function",
+              "image": "ghcr.io/husseinyoussef/afd-log-forwarder:1.2",
```
ibradwan, line 129: We shouldn't user aliases here.

devops/templates/ai-visibility/forward-afd-traffic/azuredeploy.json
```diff
@@ -0,0 +1,300 @@
...
+                  "metadata": {
+                    "timezone": "UTC",
+                    "start": "15 */12 * * *",
+                    "end": "0 */12 * * *",
+                    "desiredReplicas": "2"
+                  }
+                }
+              },
+              {
+                "name": "cpu-scaling",
```
ibradwan: Can we leave the whole scaling rules up to the platform?

devops/templates/ai-visibility/forward-afd-traffic/azuredeploy.json
```diff
@@ -0,0 +1,300 @@
...
+                }
+              ]
+            }
+          ],
+          "scale": {
+            "minReplicas": 0,
+            "maxReplicas": 10,
+            "rules": [
+              {
+                "name": "active-window",
```
ibradwan, line 171: Where's the rule the forces it to go to zero every 12 hours?

devops/templates/ai-visibility/forward-afd-traffic/createUiDefinition.json
```diff
@@ -0,0 +1,89 @@
...
+            "location": "all"
+          }
+        }
+      },
+      {
+        "name": "projectId",
+        "type": "Microsoft.Common.TextBox",
+        "label": "Clarity Project ID",
+        "toolTip": "Your Clarity project ID where bot detection data will be sent.",
+        "placeholder": "abc123xyz",
```
ibradwan: Let's not suggest anything here.

devops/templates/ai-visibility/forward-afd-traffic/createUiDefinition.json
```diff
@@ -0,0 +1,89 @@
...
+        "constraints": {
+          "required": true,
+          "validations": [
+            {
+              "regex": "^.+$",
```
ibradwan, line 74: What would happen if the user entered an incorrect project id, then tried to re-integrate?

ibradwan: What would be the best option for the user to fix this issue?

ibradwan, line 74: How does a user opt-out of sending the traffic logs to us?

devops/templates/ai-visibility/forward-afd-traffic/createUiDefinition.json
```diff
@@ -0,0 +1,89 @@
...
+              "message": "Project ID is required."
```
ibradwan, line 75: What would happen if a user tries to re-deploy the same template again?

devops/templates/ai-visibility/forward-afd-traffic/createUiDefinition.json
```diff
@@ -0,0 +1,89 @@
+{
+  "$schema": "https://schema.management.azure.com/schemas/0.1.2-preview/CreateUIDefinition.MultiVm.json#",
```
ibradwan, line 2: Any field to set in order to name the template with a familiar name?

devops/templates/ai-visibility/forward-afd-traffic/createUiDefinition.json
```diff
@@ -0,0 +1,89 @@
...
+            "subscription": "onBasics",
+            "location": "all"
+          }
+        }
+      },
+      {
+        "name": "projectId",
+        "type": "Microsoft.Common.TextBox",
+        "label": "Clarity Project ID",
+        "toolTip": "Your Clarity project ID where bot detection data will be sent.",
```
ibradwan, line 69: Have we found a way to auto populate the project id?

devops/templates/ai-visibility/forward-afd-traffic/createUiDefinition.json
```diff
@@ -0,0 +1,89 @@
...
+          "filter": {
+            "subscription": "onBasics",
+            "location": "all"
+          }
+        }
+      },
+      {
+        "name": "projectId",
+        "type": "Microsoft.Common.TextBox",
+        "label": "Clarity Project ID",
```
ibradwan, line 68: Any possible way to generate a token using 1st part AADs?

devops/templates/ai-visibility/forward-afd-traffic/eventhub-function.bicep
```diff
@@ -0,0 +1,257 @@
...
+    tier: 'Standard'
+    capacity: 1
+  }
+}
+
+resource eventHub 'Microsoft.EventHub/namespaces/eventhubs@2023-01-01-preview' = {
+  parent: ehNamespace
+  name: 'afd-logs'
+  properties: {
+    partitionCount: 2
```
ibradwan: What would be better: partitions > AzFunc instances or the other way around? We need to check from cost and perf perspectives. AFAIR, having the partition count equals or more than the consumers count is generally more efficient; however, cannot recall what the cost implications of increasing the partitions are.

devops/templates/ai-visibility/forward-afd-traffic/eventhub-function.bicep
```diff
@@ -0,0 +1,257 @@
...
+          name: 'eventhub-connection'
+          value: listKeys('${ehNamespace.id}/authorizationRules/RootManageSharedAccessKey', ehNamespace.apiVersion).primaryConnectionString
+        }
+      ]
+    }
+    template: {
+      containers: [
+        {
+          name: 'function'
+          image: 'ghcr.io/husseinyoussef/afd-log-forwarder:1.2'
```
ibradwan: As I mush as I appreciate your work, we shouldn't have any aliases in the url 😄

devops/templates/ai-visibility/forward-afd-traffic/eventhub-function.bicep
```diff
@@ -0,0 +1,257 @@
...
+            {
+              name: 'EVENT_HUB_CONSUMER_GROUP'
+              value: consumerGroup.name
+            }
+            {
+              name: 'EVENT_HUB_CONNECTION'
+              secretRef: 'eventhub-connection'
+            }
+            {
+              name: 'TARGET_HTTP_ENDPOINT'
```
ibradwan, line 171: We probably need to add the App Insights configs as empty configs here, this would be a hint for the users that they can plug in the configs and the service would start dumping logs.

devops/templates/ai-visibility/forward-afd-traffic/eventhub-function.bicep
```diff
@@ -0,0 +1,257 @@
...
+              name: 'EVENT_HUB_NAME'
+              value: eventHub.name
+            }
+            {
+              name: 'EVENT_HUB_CONSUMER_GROUP'
+              value: consumerGroup.name
+            }
+            {
+              name: 'EVENT_HUB_CONNECTION'
+              secretRef: 'eventhub-connection'
```
ibradwan, line 168: Is using the connection string fine? Or should we use managed identities instead?

devops/templates/ai-visibility/forward-afd-traffic/createUiDefinition.json
HusseinYoussef: It will forward the logs to an incorrect project then.

The project-id along with the endpoint are set as env-variables in the container app, he can update it manually if needed

image

We can show in the dashboard instructions the project id along with a copy-to-clipboard button to avoid such scenarios if there was no way to pre-populate it.

devops/templates/ai-visibility/forward-afd-traffic/createUiDefinition.json
HusseinYoussef: It will fail as we are using the same resource names.

I remember we discussed this and we decided to go with this to fail deliberately. Alternatively, we can use a random id with resource names to ensure uniqueness.

devops/templates/ai-visibility/forward-afd-traffic/azuredeploy.json
HusseinYoussef: It is this cron scheduler
image

Every 12 hours, it goes into a cooldown period (5 mins) it then scales down to 0 denoted by the minReplicas here, then at min 15, it scales backup to 2 replicas, and based on the CPU usage, it can go up to 10 replicas denoted by the maxReplicas

devops/templates/ai-visibility/forward-afd-traffic/azuredeploy.json
HusseinYoussef: No, we need to define triggers for the scaling (http requests, CPU, Cron, events, ...)
But I am thinking now that if CPU utilization is consistently > 80%, we won't be able to scale down to 0 to pull the latest image, that's what having multiple scaling rules do. So, we should rely only on the cron scheduler

ibradwan: 80% average is a bit high... let's make it 50% on average to scale out.

devops/templates/ai-visibility/forward-afd-traffic/createUiDefinition.json
HusseinYoussef: This should be as simple as deleting the diagnostic rule in the AFD. This should stop forwarding the logs into the event hub

image

devops/templates/ai-visibility/forward-afd-traffic/eventhub-function.bicep
HusseinYoussef: Yes, it should be fine, they are set as secrets in the container app, they can't be viewed. Identities could be used but it will complicate things further with roles and assignments.

image

devops/templates/ai-visibility/forward-afd-traffic/eventhub-function.bicep
HusseinYoussef: I've double checked that and it is better to have number of partitions >= AzFunc instances for better performance and parallelism. The number of partitions have no direct implications on the cost, but it can increase the runtime of the AzFunc instances as more instances will work in parallel which will increase the cost.

So, I think having like 4 partitions and 4 replicas max should be balanced.

devops/templates/ai-visibility/forward-afd-traffic/eventhub-function.bicep
HusseinYoussef: App insights isn't supported by Container apps out-of-the box

devops/templates/ai-visibility/forward-afd-traffic/eventhub-function.bicep
HusseinYoussef: I had to use my personal account to upload the container image 😁 until I manage to push to the org instead.
I've managed to push the image to the Microsoft org now and linked it to the Clarity repo.

devops/templates/ai-visibility/forward-afd-traffic/createUiDefinition.json
HusseinYoussef: Removed it.

devops/templates/ai-visibility/forward-afd-traffic/azuredeploy.json
```diff
@@ -0,0 +1,268 @@
+{
```
nairmanu, line 1: why is this PR in this repository?

HusseinYoussef:
Closing this as it is moved to Clarity Dashboard repo.

PR #1068, opened 2026-03-29
Optimizations for smaller minified output

nairmanu:
## Commit 1: `build: optimize terser config and rollup output options`

- **`passes: 3`** — Runs terser compression 3 times. Each pass finds new optimization opportunities from the previous pass's output (e.g., inlining single-call-site functions, then eliminating the dead variables left behind).
- **`hoist_vars: true`** — Consolidates 379 scattered `var` declarations into ~164 (mostly one declaration at the IIFE top). Costs 266 raw bytes for the preamble but saves ~98 gzip bytes because gzip compresses a single `var t,e,n,...` block better than hundreds of scattered `var` keywords.
- **`freeze: false`** on all 7 IIFE outputs — Removes one `Object.freeze()` wrapper around the Clarity public API namespace object that Rollup generates. Saves ~15 raw bytes.
- **Shared `terserOpts` config** — Extracted repeated `terser({output: {comments: false}})` into a shared constant applied to all IIFE builds.

**Options evaluated and intentionally excluded:**
- **`toplevel: true`** — No-op in an IIFE (everything is already inside a function scope).
- **`unsafe_comps: true`** — Flips comparisons (`a < b` → `b > a`). Saves raw bytes but hurts gzip by +5 bytes because reversed comparisons create less common byte patterns.
- **`pure_getters: true`** — Assumes all property access has no side effects. Enables variable inlining (−70 gzip). Excluded because it's a global declaration that could conflict with Zone.js proxy behavior, browser extensions wrapping DOM objects, or future code adding getters.
- **`module: true`** — Tells terser the code is an ES module. Removes `"use strict"`. Excluded because the IIFE is not an ES module — removing strict mode changes runtime semantics (`this` in plain function calls becomes `window` instead of `undefined`).

## Commit 2: `refactor: eliminate rollup namespace objects via named imports`

- Replaces `import * as X from "..."` with named imports `{ start, stop }` where modules are only used in `modules[]` arrays for lifecycle management.

- When Rollup sees `import * as X`, it creates a runtime namespace object containing all exports from that module — even if only `start` and `stop` are used. By switching to `import { start, stop }`, Rollup can tree-shake the namespace and only include what's needed.

- `data/index.ts` was the biggest win: 13 sub-module namespaces (`baseline`, `dimension`, `variable`, `limit`, `summary`, `cookie`, `consent`, `metadata`, `envelope`, `upload`, `ping`, `upgrade`, `extract`) were converted from full namespace objects to minimal `{start, stop}` pairs. The `compute()` calls that previously went through the namespace were converted to direct named function imports.

- `layout/node.ts` had a special case: `dom[call]` used dynamic property access on the `dom` namespace, which forced Rollup to keep the entire namespace as a runtime object. Replaced with a `domCall()` helper using explicit if/else dispatch on a boolean flag, allowing Rollup to inline everything.

## Commit 3: `refactor: deduplicate repeated code patterns and simplify modules`

- **`baseline.ts`** — Used `Object.assign` for baseline state copy instead of 20+ individual property assignments. Extracted `updatePointer` helper to deduplicate 3 identical pointer-update blocks.
- **`mutation.ts`** — Extracted `proxyRule` helper to deduplicate 4 nearly identical CSS rule proxy blocks (`insertRule`/`deleteRule` on `CSSStyleSheet` and `CSSMediaRule`).
- **`style.ts`** — Extracted `proxyStyleMethod` helper for `CSSStyleSheet` method proxying. Simplified `arraysEqual` comparison.
- **`history.ts`** — Extracted `proxyHistory` helper to deduplicate `pushState`/`replaceState` proxy setup.
- **`animation.ts`** — Simplified `overrideAnimationHelper` to iterate over names array. Removed 5 unused module-level variables.
- **`metric.ts`** — Extracted `init()` helper to deduplicate metric initialization checks across `count`, `sum`, and `max`.
- **`compress.ts`** — Replaced manual `read()` stream consumer loop (10 lines) with `new Response(stream).arrayBuffer()` (1 line). Same browser support requirements, eliminates an entire function.
- **`signal.ts`** — Eliminated redundant `parseSignals()` wrapper function and simplified inner try/catch.
- **Shortened internal property names** — `clarityOverrides` → `__clr` (20 occurrences), `clarityAnimationId` → `__clrAId`, `clarityOperationCount` → `__clrOCnt`, `claritySheetId` → `__clrSId`. These are internal properties attached to DOM elements/window, not part of any public API.

## Commit 4: `refactor: modernize syntax for smaller minified output`

- **`indexOf(x) >= 0` → `includes(x)`** — ~27 occurrences across 15+ files. `includes()` produces shorter minified output than the `indexOf` + comparison pattern. The codebase already uses `includes()` in existing code, so this doesn't change browser support requirements. Both use strict-equality-like comparison (`SameValueZero` vs `===`); the only difference is `NaN` handling, which is irrelevant for string/tag comparisons.
- **Template literals → `+` concatenation** — In files like `selector.ts`, `encode.ts`, `scrub.ts`, `dom.ts`. When targeting ES5, TypeScript compiles template literals to `.concat()` chains which are verbose. Plain `+` concatenation produces shorter output.
- **Consolidated `tokens.push()`** — In `diagnostic/encode.ts`, `interaction/encode.ts`, `layout/encode.ts`, `performance/encode.ts`. Sequential `tokens.push(a); tokens.push(b); tokens.push(c)` → `tokens.push(a, b, c)`. Eliminates repeated method lookups.
- **`.toString()` → `"" + x`** — 6 occurrences in `metadata.ts`, `consent.ts`, `observer.ts`, `data/encode.ts`. `""+x` minifies to 4 characters vs `.toString()` which stays at 12.
- **Inlined `str()` helper** — Single-use wrapper function in `layout/encode.ts` eliminated.
- **Local alias for `b.data`** — In `data/encode.ts`, aliased deeply nested `b.data` to a local variable `d` in the baseline encoding case, eliminating 25 repeated `.data.` property chains.


packages/clarity-js/src/layout/style.ts
```diff
@@ -21,39 +21,26 @@ function proxyStyleRules(win: any) {
...
-                    // if we haven't seen this stylesheet on this page yet, wait until the checkDocumentStyles has found it
-                    // and attached the sheet to a document. This way the timestamp of the style sheet creation will align
-                    // to when it is used in the document rather than potentially being misaligned during the traverse process.
-                    if (createdSheetIds.indexOf(this[styleSheetId]) > -1) {
-                        trackStyleChange(time(), this[styleSheetId], StyleSheetOperation.ReplaceSync, arguments[0]);
-                    }                
+function proxyStyleMethod(win: any, method: string, operation: StyleSheetOperation): void {
+    if (win.__clr[method] === undefined) {
+        win.__clr[method] = win.CSSStyleSheet.prototype[method];
+        win.CSSStyleSheet.prototype[method] = function(): any {
```
malaref: no need for the `: any`

packages/clarity-js/src/layout/style.ts
```diff
@@ -21,39 +21,26 @@ function proxyStyleRules(win: any) {
...
+    win.__clr = win.__clr || {};
 
     if (win['CSSStyleSheet'] && win.CSSStyleSheet.prototype) {
-        if (win.clarityOverrides.replace === undefined) { 
-            win.clarityOverrides.replace = win.CSSStyleSheet.prototype.replace; 
-            win.CSSStyleSheet.prototype.replace = function(): Promise<CSSStyleSheet> {
-                if (core.active()) {
-                    // if we haven't seen this stylesheet on this page yet, wait until the checkDocumentStyles has found it
-                    // and attached the sheet to a document. This way the timestamp of the style sheet creation will align
-                    // to when it is used in the document rather than potentially being misaligned during the traverse process.
```
malaref, line 33: these comments shouldn't be omitted; they don't make it to the built bundle anyway

packages/clarity-js/src/layout/node.ts
```diff
@@ -14,11 +17,15 @@ import { electron } from "@src/data/metadata";
 const IGNORE_ATTRIBUTES = ["title", "alt", "onload", "onfocus", "onerror", "data-drupal-form-submit-last", "aria-label"];
 const newlineRegex = /[\r\n]+/g;
 
+function domCall(isAdd: boolean, node: Node, parent: Node, data: any, source: Source): void {
+    isAdd ? domAdd(node, parent, data, source) : domUpdate(node, parent, data, source);
```
malaref: it'd be better if the function is just passed as the param instead of this boolean flag

packages/clarity-js/src/layout/mutation.ts
```diff
@@ -346,67 +346,40 @@ function proxyStyleRules(win: any): void {
     return;
   }
 
-  win.clarityOverrides = win.clarityOverrides || {};
+  win.__clr = win.__clr || {};
 
-  // Some popular open source libraries, like styled-components, optimize performance
-  // by injecting CSS using insertRule API vs. appending text node. A side effect of
-  // using javascript API is that it doesn't trigger DOM mutation and therefore we
-  // need to override the insertRule API and listen for changes manually.
```
malaref, line 354: the re-written comment omits key info

malaref (approved):

Niyibitanga (approved):

PR #1088, opened 2026-05-20
Add extended region tracking with viewport enter/exit events (0.8.65)

Niyibitanga:
## Summary

Introduces v2 implementations of region, target, and encode modules, swapped in for the extended build via a Rollup alias. The standard build wire format is unchanged.

- **region-v2** — tracks viewport enter/exit via IntersectionObserver in addition to the existing Rendered/Visible/ScrolledToEnd lifecycle. Unlike v1, observation continues past `ScrolledToEnd` since enter/exit needs a continuous visibility signal.
- **target-v2** — populates region id on `TargetMetadata` so interaction events (pointer/click/input/scroll) carry their region attribution.
- **encode-v2** — unconditional region push on all events, fixing the mouse-pointer region gap caused by `id`/`isPrimary` being optional.
- **encode (standard)** — pointer events now always emit `id` and `isPrimary` in fixed positions, using `Constant.Empty` as the absent-`id` sentinel and `"true"` as the absent-`isPrimary` default. Matches the backend's `ValueKind`-guarded fallback in `PayloadDecoder.cs`, so no backend change required.
- **clarity-decode** — pointer decoder reads `id`/`isPrimary` positionally with type-guarded fallbacks; backward compatible with 5/6/7-token legacy payload shapes.

## Note on `RegionVisibility` enum values

Values are intentionally non-contiguous (`0, 10, 13, 14, 15`). The `Rendered`/`Visible`/`ScrolledToEnd` values were set in 2021 so legacy single-enum payloads could be disambiguated from `InteractionState` (≥16) by numeric range — values ≤13 route to visibility, ≥16 route to interaction. `ViewportEnter`/`ViewportExit` are placed at 14/15 to stay within the visibility range and preserve that disambiguation, rather than colliding with `InteractionState.Clicked = 20`.

### DIFF
--- a/packages/clarity-js/rollup.config.ts
+++ b/packages/clarity-js/rollup.config.ts
@@ -54,6 +54,12 @@ export default [
       warn(message);
     },
     plugins: [
+      alias({
+        entries: [
+          { find: '@src/layout/region', replacement: '@src/layout/regionV2' },
+          { find: '@src/interaction/encode', replacement: '@src/interaction/encodeV2' }
+        ]
+      }),
       resolve(),
       typescript(),
       terser(terserOpts),
--- a/packages/clarity-js/src/layout/target.ts
+++ b/packages/clarity-js/src/layout/target.ts
@@ -15,7 +15,7 @@ export function target(evt: UIEvent): Node {
 
 export function metadata(node: Node, event: Event, text: string = null): TargetMetadata {
     // If the node is null, we return a reserved value for id: 0. Valid assignment of id begins from 1+.
-    let output: TargetMetadata = { id: 0, hash: null, privacy: Privacy.Text };
+    let output: TargetMetadata = { id: 0, hash: null, privacy: Privacy.Text, region: -1 };
     if (node) {
         let value = dom.get(node);
         if (value !== null) {
@@ -23,6 +23,7 @@ export function metadata(node: Node, event: Event, text: string = null): TargetM
             output.id = value.id;
             output.hash = value.hash;
             output.privacy = metadata.privacy;
+            output.region = value.region || -1;
             if (value.region) { region.track(value.region, event); }
             if (metadata.fraud) { fraud.check(metadata.fraud, value.id, text || value.data.value); }
         }
--- a/packages/clarity-js/types/layout.d.ts
+++ b/packages/clarity-js/types/layout.d.ts
@@ -53,7 +53,9 @@ export interface CustomElementData {
 export const enum RegionVisibility {
     Rendered = 0,
     Visible = 10,
-    ScrolledToEnd = 13
+    ScrolledToEnd = 13,
+    ViewportEnter = 14,
+    ViewportExit = 15
 }
 
 export const enum Constant {
@@ -260,6 +262,7 @@ export interface TargetMetadata {
     id: number;
     hash: [string, string];
     privacy: Privacy;
+    region?: number;
 }
 
 export interface StyleSheetData {
--- a/lerna.json
+++ b/lerna.json
@@ -2,6 +2,6 @@
   "packages": [
     "packages/*"
   ],
-  "version": "0.8.64",
+  "version": "0.8.65",
   "npmClient": "yarn"
 }
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "clarity",
   "private": true,
-  "version": "0.8.64",
+  "version": "0.8.65",
   "repository": "https://github.com/microsoft/clarity.git",
   "author": "Sarvesh Nagpal <sarveshn@microsoft.com>",
   "license": "MIT",
--- a/packages/clarity-decode/package.json
+++ b/packages/clarity-decode/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clarity-decode",
-  "version": "0.8.64",
+  "version": "0.8.65",
   "description": "An analytics library that uses web page interactions to generate aggregated insights",
   "author": "Microsoft Corp.",
   "license": "MIT",
@@ -26,7 +26,7 @@
     "url": "https://github.com/Microsoft/clarity/issues"
   },
   "dependencies": {
-    "clarity-js": "^0.8.64"
+    "clarity-js": "^0.8.65"
   },
   "devDependencies": {
     "@rollup/plugin-commonjs": "^24.0.0",
--- a/packages/clarity-decode/src/interaction.ts
+++ b/packages/clarity-decode/src/interaction.ts
@@ -18,7 +18,8 @@ export function decode(tokens: Data.Token[]): InteractionEvent {
                 target: tokens[2] as number,
                 x: tokens[3] as number,
                 y: tokens[4] as number,
-                id: tokens[5] as number | undefined,
+                id: typeof tokens[5] === "number" ? tokens[5] as number : undefined,
+                isPrimary: tokens[6] === undefined ? undefined : tokens[6] === "true",
             };
             return { time, event, data: pointerData };
         case Data.Event.Click:
--- a/packages/clarity-devtools/package.json
+++ b/packages/clarity-devtools/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clarity-devtools",
-  "version": "0.8.64",
+  "version": "0.8.65",
   "private": true,
   "description": "Adds Clarity debugging support to browser devtools",
   "author": "Microsoft Corp.",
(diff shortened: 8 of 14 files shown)


packages/clarity-js/src/insight/snapshot.ts
```diff
@@ -33,7 +33,7 @@ export function target(evt: UIEvent): Node {
 }
 
 export function metadata(node: Node): TargetMetadata {
-    let output: TargetMetadata = { id: 0, hash: null, privacy: config.conversions ? Privacy.Sensitive : Privacy.Snapshot };
+    let output: TargetMetadata = { id: 0, hash: null, privacy: config.conversions ? Privacy.Sensitive : Privacy.Snapshot, region: 0 };
```
nairmanu: remove. if default is required, use null or -1

packages/clarity-js/src/interaction/encode.ts
```diff
@@ -38,12 +38,13 @@ export default async function (type: Event, ts: number = null): Promise<void> {
...
                     tokens.push(entry.data.y);
-                    if (entry.data.id !== undefined) { 
-                        tokens.push(entry.data.id); 
+                    if (entry.data.id !== undefined) {
+                        tokens.push(entry.data.id);
 
                         if (entry.data.isPrimary !== undefined) {
                             tokens.push("" + entry.data.isPrimary);
                         }
+                        if (pTarget.region) { tokens.push(pTarget.region); }
```
nairmanu: should not be gated

packages/clarity-js/src/interaction/encode.ts
```diff
@@ -69,6 +70,7 @@ export default async function (type: Event, ts: number = null): Promise<void> {
                     entry.data.w, entry.data.h,
                     entry.data.tag, entry.data.class, entry.data.id, entry.data.source
                 );
+                if (cTarget.region) { tokens.push(cTarget.region); }
```
nairmanu: do not gate

packages/clarity-js/src/layout/target.ts
```diff
@@ -15,14 +16,15 @@ export function target(evt: UIEvent): Node {
 
 export function metadata(node: Node, event: Event, text: string = null): TargetMetadata {
     // If the node is null, we return a reserved value for id: 0. Valid assignment of id begins from 1+.
-    let output: TargetMetadata = { id: 0, hash: null, privacy: Privacy.Text };
+    let output: TargetMetadata = { id: 0, hash: null, privacy: Privacy.Text, region: 0 };
```
nairmanu: dont use 0

packages/clarity-js/src/layout/target.ts
```diff
@@ -15,14 +16,15 @@ export function target(evt: UIEvent): Node {
...
     if (node) {
         let value = dom.get(node);
         if (value !== null) {
             let metadata = value.metadata;
             output.id = value.id;
             output.hash = value.hash;
             output.privacy = metadata.privacy;
+            output.region = extended.getRegion(value);
```
nairmanu: output.region = value.region

packages/clarity-js/types/layout.d.ts
```diff
@@ -53,7 +53,9 @@ export interface CustomElementData {
 export const enum RegionVisibility {
     Rendered = 0,
     Visible = 10,
-    ScrolledToEnd = 13
+    ScrolledToEnd = 13,
+    ViewportEnter = 20,
```
nairmanu: why 20 and 21

packages/clarity-js/types/layout.d.ts
```diff
@@ -260,6 +262,7 @@ export interface TargetMetadata {
     id: number;
     hash: [string, string];
     privacy: Privacy;
+    region: number;
```
nairmanu: null | number

packages/clarity-js/rollup.config.ts
```diff
@@ -23,6 +23,11 @@ export default [
       { file: pkg.module, format: "es", exports: "named" }
     ],
     plugins: [
+      alias({
+        entries: [
+          { find: '@src/layout/region-extended', replacement: '@src/layout/region-extended-blank' }
```
nairmanu: layout/region replace:layout/revionV2
interaction/encode replace:interaction/encodeV2

packages/clarity-js/src/layout/region-extended.ts
```diff
@@ -0,0 +1,30 @@
+import { RegionData, RegionVisibility } from "@clarity-types/layout";
+import { NodeValue } from "@clarity-types/layout";
+
+let wasIntersecting: WeakMap<Element, boolean> = null;
```
nairmanu: check if its Node vs Element

nairmanu: nit:better name than wasIntersecting

packages/clarity-js/src/insight/snapshot.ts
Niyibitanga: Addressed by the v2 refactor — this file was reverted to master or deleted. See consolidated comment above.

Niyibitanga:
Thanks for the review! We refactored the approach since your review — instead of modifying existing files with conditional gating, we now use v2 file copies (`region-v2.ts`, `encode-v2.ts`) swapped via Rollup alias for the extended build only. Standard build files are unchanged from master except `target.ts` (minor: sets `region` on TargetMetadata, harmless — never pushed to tokens in standard build). `region-extended.ts` and `region-extended-blank.ts` were deleted. `target-v2.ts` was also eliminated.

Your feedback directly influenced this refactor:
- Region push is now unconditional in `encode-v2.ts` (no gating) ✅
- Region is pushed outside the `id` gate — mouse events get 6-token format `[time, event, target, x, y, region]` ✅
- `snapshot.ts` unchanged from master ✅
- `TargetMetadata.region` is optional (`region?: number`) ✅
- `region-v2.ts` uses `WeakMap<Element, boolean>` (not Node) ✅
- Renamed `wasIntersecting` → `prevIntersected` ✅

Please re-review on the updated diff — only 4 files changed from master (down from 8).

packages/clarity-js/src/interaction/encode.ts
Niyibitanga: Addressed by the v2 refactor — this file was reverted to master or deleted. See consolidated comment above.

packages/clarity-js/src/interaction/encode.ts
Niyibitanga: Addressed by the v2 refactor — this file was reverted to master or deleted. See consolidated comment above.

packages/clarity-js/src/layout/target.ts
Niyibitanga: Addressed by the v2 refactor — this file was reverted to master or deleted. See consolidated comment above.

packages/clarity-js/src/layout/target.ts
Niyibitanga: Addressed by the v2 refactor — this file was reverted to master or deleted. See consolidated comment above.

packages/clarity-js/types/layout.d.ts
Niyibitanga: Follows the existing RegionVisibility numbering pattern with gaps for extensibility: Rendered=0, Visible=10, ScrolledToEnd=13. Values 20/21 leave room for additional visibility states without renumbering.

packages/clarity-js/types/layout.d.ts
Niyibitanga: Addressed by the v2 refactor — this file was reverted to master or deleted. See consolidated comment above.

packages/clarity-js/rollup.config.ts
Niyibitanga: Addressed by the v2 refactor — this file was reverted to master or deleted. See consolidated comment above.

packages/clarity-js/src/layout/region-extended.ts
Niyibitanga: We used WeakMap<Element, boolean> here intentionally — IntersectionObserverEntry.target is typed as Element in the DOM API (not Node), so Element is the      more precise type for the intersection state map. Text nodes and other non-Element nodes can never appear as IntersectionObserver targets.                  

  That said, regionMap uses WeakMap<Node, string> and tracks the same set of objects, so there's an inconsistency. Would you prefer we align prevIntersected  
  to Node for consistency with regionMap, or change both maps to Element since in practice only Elements are ever observed (both via querySelectorAll and the 
  nodeType === ELEMENT_NODE guard in observe())?

Niyibitanga: Renamed to `prevIntersected` in `region-v2.ts`.

nairmanu (approved):

jsj85 (approved):

PR #1095, opened 2026-06-02
Log GPC flag and enable diagnostics

shawj0:


packages/clarity-js/src/clarity.ts
```diff
@@ -28,9 +28,9 @@ const modules: Module[] = [
 ];
 
 export function start(config: Config = null): void {
+  core.config(config);
```
nairmanu: why is this moved?

packages/clarity-js/types/data.d.ts
```diff
@@ -394,7 +395,7 @@ export const enum ConsentSource {
     APIv2 = 5,
     Cookie = 6,
     Default = 7,
-    // 100-255 Reserved for CMP integration, both internal and external
+    // 100-199 Reserved for CMP integration, both internal and external
```
nairmanu: why is this updated?

packages/clarity-js/types/data.d.ts
```diff
@@ -409,6 +410,7 @@ export const enum ConsentSource {
     CmpWebTofee = 156,
     CmpWPConsent = 157,
     CmpSeersAI = 158,
+    // 252-254 reserved for internal use
```
nairmanu: remove changes in this file

packages/clarity-js/src/clarity.ts
shawj0: Config needs to be applied before check() so that user-provided overrides are available during capability evaluation

packages/clarity-js/src/data/metadata.ts
```diff
@@ -51,6 +51,7 @@ export function start(): void {
   dimension.log(Dimension.AncestorOrigins, ancestorOrigins);
   dimension.log(Dimension.Timezone, timezone);
   dimension.log(Dimension.TimezoneOffset, timezoneOffset);
+  dimension.log(Dimension.GlobalPrivacyControl, "" + !!(navigator && navigator.globalPrivacyControl));
```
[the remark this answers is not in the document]
shawj0: resolved

packages/clarity-js/src/data/metadata.ts
```diff
@@ -51,6 +51,7 @@ export function start(): void {
   dimension.log(Dimension.AncestorOrigins, ancestorOrigins);
   dimension.log(Dimension.Timezone, timezone);
   dimension.log(Dimension.TimezoneOffset, timezoneOffset);
+  dimension.log(38 as any, "" + !!(navigator && navigator["globalPrivacyControl"]));
```
nairmanu: use enum instead of magic number 38

packages/clarity-js/src/core/index.ts
```diff
@@ -43,7 +43,7 @@ export function check(): boolean {
             "now" in Date &&
             "now" in performance &&
             typeof WeakMap !== "undefined" &&
-            !globalPrivacyControlSet
+            (!globalPrivacyControlSet || configuration.diagnostics)
```
nairmanu: i dont think configuration holds the vlaues we want. you will have to pass config to check()

packages/clarity-js/src/clarity.ts
nairmanu: wouldnt it be better to pass config to core.check() instead?

shawj0: we can do that, passing config in now

packages/clarity-js/src/core/index.ts
shawj0: It has the values now that we pass it in

packages/clarity-js/src/data/metadata.ts
shawj0: using the enum now

nairmanu (approved):

jsj85 (approved):

PR #1104, opened 2026-06-26
fix(clarity-js): prevent history proxy stack overflow on Chrome for iOS (CriOS)

nairmanu:
Fixes # `RangeError: Maximum call stack size exceeded` thrown only on Chrome for iOS (CriOS).1103 

**Cause:** `core/history.ts` proxies `history.pushState`/`replaceState`. CriOS injects its own history wrapper; when it re-enters Clarity's proxy the two wrappers call each other until the stack overflows. The existing `CallStackDepth` guard only gated `compute()` (and reset its counter every navigation), so the real re-entry  `original. was never protected.apply` point 

**Fix:** track re-entrancy `depth` around `original.apply`. Normal navigations are unchanged; once depth reaches `Setting.CallStackDepth` the proxy stops calling `original` and logs `Code.CallStackDepth`, so the stack can't overflow. Removes the now-redundant `check()`/`count`.

**Testing:** `yarn build` passes; `yarn tslint` adds no new findings in `history.ts`.

### DIFF
--- a/packages/clarity-js/src/core/history.ts
+++ b/packages/clarity-js/src/core/history.ts
@@ -8,11 +8,11 @@ import * as metric from "@src/data/metric";
 let pushState = null;
 let replaceState = null;
 let url = null;
-let count = 0;
+let depth = 0;
 
 export function start(): void {
     url = getCurrentUrl();
-    count = 0;
+    depth = 0;
     bind(window, "popstate", compute);
     pushState = proxyHistory(pushState, "pushState");
     replaceState = proxyHistory(replaceState, "replaceState");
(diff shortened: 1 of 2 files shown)


rgaleev:
Can we add a test to validate the behavior?

nairmanu:
Added an e2e test in `test/history.test.ts` (pushed in ebb4671).

It reproduces the CriOS scenario — a foreign wrapper that re-dispatches through the live `history.pushState`, so once Clarity wraps it the two wrappers call each other — and asserts that:

1. the page does **not** throw `RangeError: Maximum call stack size exceeded`, and
2. Clarity emits a `Code.CallStackDepth` diagnostic log instead.

I confirmed it **fails against the previous unguarded implementation** (the trigger throws a `RangeError`) and passes with the fix.

rgaleev (approved):

Niyibitanga (approved):

PR #1109, opened 2026-06-30
Assign a stable tab id when consent is granted

Niyibitanga:


packages/clarity-js/src/data/metadata.ts
```diff
@@ -197,6 +197,7 @@ export function consentv2(consentState: ConsentState = defaultStatus, source: nu
   if (core.active() && consentData.analytics_Storage) {
     config.track = true;
     track(user(), BooleanFlag.True);
+    dimension.log(Dimension.TabId, tab());
```
nairmanu, line 200: i dont think this is the correct approach. a tab id would be assigned at the beginning of the session - irrespective of consent. it will be persisted only if consent is granted. here you will be creating a new tab id. what we should be doing is persist already generated tabid in session storage if consent is granted

nairmanu (changes requested):

nairmanu (approved):

packages/clarity-js/src/data/metadata.ts
Niyibitanga: tab id is a last dimension - new id will override the old one.

rgaleev (approved):

PR #1115, opened 2026-07-06
Fix consent API failing to delete domain-scoped cookies

### LINKED ISSUE #1105: Consent API fails to delete domain-scoped cookies when internal cookie domain cache is uninitialized
**Bug description**

Calling either

window.clarity("consentv2", {
    analytics_Storage: "denied",
    ad_Storage: "denied"
});

or

window.clarity("consent", false);

does not remove existing Clarity cookies when running on a subdomain (for example *.azurewebsites.net) if the internal cookie domain cache has not yet been initialized.

The issue does not reproduce on localhost.

During debugging, both APIs execute the expected consent path (Kr -> Qr -> ma), but the domain discovery logic falls back to an empty cached domain. The subsequent _clck deletion is then attempted without a domain attribute, so the existing domain-scoped _clck cookie remains.

If the internal cached domain has already been initialized, cookie deletion succeeds.

In our case, this occurs when Clarity is used without OneTrust being loaded on the page, so no previous Clarity command appears to initialize the cookie domain cache before consent is revoked.

**Repro Steps**

1. Load Clarity on a subdomain (for example *.azurewebsites.net).
2. Ensure existing Clarity cookies are already present (_clck and _clsk), but the internal cookie domain cache has not been initialized in the current Clarity runtime.
3. Call

window.clarity("consentv2", {
    analytics_Storage: "denied",
    ad_Storage: "denied"
});

(or window.clarity("consent", false)).

4. Observe that _clck remains while _clsk is recreated.

The same sequence on localhost removes both cookies correctly.

**Expected behavior**

Both _clck and _clsk should be removed when consent is revoked, regardless of whether the internal cookie domain cache has previously been initialized.

**Actual behavior**

The consent API reaches the cookie deletion code path, but _clck remains because the delete is issued without the correct domain attribute.

**Debugging notes**

The following execution path is observed:

Kr(...)
    -> Qr(true)
        -> ma("_clsk", "", 0)
        -> ma("_clck", "", 0)
        -> ma("_clsk", "<value>", 1)

The first ma() call performs cookie domain discovery.

The discovery attempts:

.net
.azurewebsites.net
.subdomain.azurewebsites.net

but eventually falls back to:

_ = ""

The subsequent _clck deletion therefore executes without a domain attribute.

Manual browser testing confirms the browser behaves correctly:

document.cookie = "...;path=/"
does NOT delete the cookie.

document.cookie = "...;path=/;domain=.subdomain.azurewebsites.net"
does delete the cookie.

This suggests the fallback path after failed domain discovery is responsible.

**Additional timing observation**

The failure appears to depend on when the consent API is called.

If I call `consentv2`/`consent(false)` immediately after Clarity loads, before Clarity has performed its own normal `_clsk` session update, the internal cookie-domain cache is still uninitialized and deletion fails.

If I wait until Clarity performs its normal `_clsk` update first, the internal cookie-domain cache is initialized and the same consent API call removes the cookies correctly.

So the issue appears to occur when the first `ma()` call in the current Clarity runtime is a deletion call rather than a normal cookie write/update.

**Project info**

- Website hosted on Azure App Service (*.azurewebsites.net)

**Browser info:**

 - OS: Windows
 - Browser: Chrome 149

rgaleev:
Updated the approach based on a compliance concern with the previous probe-based discovery.
Why the change: the earlier version discovered the writable root domain by writing a throwaway  _clcp  probe cookie. But domain discovery runs during cookie deletion too (consent revoke →  clear(true)  →  setCookie(_clck, "", 0) ), which means it would write a cookie while consent is denied. Under the ePrivacy Directive (Art. 5(3)), the regulated act is storing information on the device, independent of the cookie's content — so writing even an empty/throwaway probe without consent is not permitted.

New approach (no cookie ever written under denied consent):
• Deletions no longer discover the domain. They issue the delete (an already-expired cookie, which stores nothing) on the current host and every writable parent-domain suffix. The browser ignores domains we can't write to (e.g. public suffixes), so this reliably removes a domain-scoped  _clck / _clsk  without needing the cache — fixing the deletion-first scenario in #1105.
• Writes (only when consent is granted,  config.track === true ) keep the original readback-based discovery, unchanged.
This drops the  _clcp  constant and the probe entirely.
Review comments addressed in the regression test: artifact loaded in  beforeAll ,  route  handler is  async /awaits  fulfill , and the fixed  setTimeout  is replaced with  expect.poll .

Fixes #1105

### DIFF
--- a/packages/clarity-js/src/data/cookie.ts
+++ b/packages/clarity-js/src/data/cookie.ts
@@ -57,8 +57,35 @@ export function setCookie(key: string, value: string, time: number): void {
     expiry.setDate(expiry.getDate() + time);
     let expires = expiry ? Constant.Expires + expiry.toUTCString() : Constant.Empty;
     let cookie = key + "=" + encodedValue + Constant.Semicolon + expires + Constant.Path;
+
+    // When clearing a cookie (empty value) we must not write any new cookie, because storing
+    // information on the device without consent is not permitted, and consent may already be denied
+    // by the time we clear. We also cannot rely on the cached root domain here: when a deletion is
+    // the first cookie operation in this runtime, the domain is still undiscovered. Instead, issue
+    // the deletion on the current host and on every writable parent domain. A deletion is an
+    // already-expired cookie, so it stores nothing and can be attempted broadly; the browser simply
+    // ignores domains we are not allowed to write to (e.g. public suffixes). This guarantees an
+    // existing domain-scoped cookie (e.g. _clck on .example.com) is removed from subdomains without
+    // needing to discover the domain first. See issue #1105.
+    if (value === Constant.Empty) {
+      // Delete the host-only cookie.
+      document.cookie = cookie;
+      let hostname = location.hostname ? location.hostname.split(Constant.Dot) : [];
+      let domain = Constant.Empty;
+      // Walk backwards, accumulating parent-domain suffixes (.com -> .example.com -> ...), and delete
+      // on each. We skip the absolute last part, e.g. .com or .net, which is never a writable domain.
+      for (let i = hostname.length - 1; i >= 0; i--) {
+        domain = Constant.Dot + hostname[i] + domain;
+        if (i < hostname.length - 1) {
+          document.cookie = cookie + Constant.Semicolon + Constant.Domain + domain;
+        }
+      }
+      return;
+    }
+
     try {
       // Attempt to get the root domain only once and fall back to writing cookie on the current domain.
+      // This only runs for writes, which happen when config.track is true (i.e. consent is granted).
       if (rootDomain === null) {
         let hostname = location.hostname ? location.hostname.split(Constant.Dot) : [];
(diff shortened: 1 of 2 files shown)


packages/clarity-js/src/data/cookie.ts
```diff
@@ -57,8 +57,35 @@ export function setCookie(key: string, value: string, time: number): void {
     expiry.setDate(expiry.getDate() + time);
     let expires = expiry ? Constant.Expires + expiry.toUTCString() : Constant.Empty;
     let cookie = key + "=" + encodedValue + Constant.Semicolon + expires + Constant.Path;
+
+    // When clearing a cookie (empty value) we must not write any new cookie, because storing
+    // information on the device without consent is not permitted, and consent may already be denied
```
Niyibitanga, line 62: NIT: comments are a bit verbose

Niyibitanga (approved):

PR #1123, opened 2026-07-20
Add standalone pointer down event

shawj0:
Adds complete legacy down-event context plus pointer pressure, width, and height as a standalone experimental wire event.

## What changed

- Adds `PointerDown = 53`.
- Wire shape: `[time, 53, target, x, y, pointerId, isPrimary, pointerType, pressure, width, height]`.
- Pointer types: Mouse=1, Touch=2, Pen=3.
- Normal and extended builds emit mouse, primary touch/pen, and non-primary touch/pen `pointerdown`.
- Carries the existing DOM target ID, rounded page-relative coordinates, browser pointer ID, and primary state.
- Preserves `PointerEvent.isPrimary` exactly as reported, including abnormal or synthetic mouse values.
- Uses the existing target metadata lookup without changing its shared API.
- Browser pressure, width, and height remain numeric without client-side rounding.
- Existing MouseDown, TouchStart, region, replay, and pointer wire shapes are unchanged.
- PointerDown uses the existing pointer state and shared legacy pointer encoder loop; its extra fields are optional for non-down events.
- Both interaction encoders use the same pointer state and shared event loop; the extended encoder keeps Event 53 on the fixed 11-token contract without adding its region token.
- clarity-decode exposes the event separately as `pointerDown`.
- Version bumped to `0.8.70`, off the beta prerelease line.

## Why

The expanded standalone contract reaches data parity with legacy down events while preserving PointerEvent-only geometry. It can be evaluated alongside the existing pointer pipeline now and can support a later, separately approved replacement of legacy MouseDown/TouchStart down records without another wire migration.

## Evidence

- clarity-js and clarity-decode builds passed.
- Pointer Playwright: 6 passed.
- Chromium, Firefox, and WebKit local smoke passed with all 11 fields.
- Mouse, primary touch, primary pen, and non-primary touch were covered.
- Current local clarity-js -> local Ingest smoke wrote the complete expanded PPG1 values, including synthetic mouse `isPrimary=false`.
- Targeted TSLint passed.
- Real-device BrowserStack was not rerun.

## Compatibility

- Old clients send no event 53 and remain unchanged.
- Old backend instances ignore unknown event 53.
- New backend instances decode event 53.
- Event 53 remains separate from legacy pointer processing, preventing duplicate trajectory, distance, baseline, activity, or down/up accounting.
- The previous event-53 shape was never published or deployed, so no migration or dual-read path is required.


packages/clarity-js/src/interaction/encode.ts
```diff
@@ -40,6 +40,12 @@ export default async function (type: Event, ts: number = null): Promise<void> {
                     tokens.push(entry.data.y);
                     tokens.push(entry.data.id !== undefined ? entry.data.id : Constant.Empty);
                     tokens.push(entry.data.isPrimary === undefined ? "true" : "" + entry.data.isPrimary);
+                    tokens.push(Constant.Empty);
+                    if (entry.event === Event.MouseDown || entry.event === Event.TouchStart) {
```
nairmanu: are the values only available for these two events?

shawj0:
WebKit follow-up validation for commit `872d915` is complete.

The implementation keeps the existing `MouseDown` / `TouchStart` records, but sources
geometry for both click and tap from `pointerdown`. TouchEvent force/radius values are
not used. A sub-pixel coordinate tolerance is used only to correlate WebKit's
`pointerdown` and `touchstart`; pressure, width, and height remain unmodified.

Local clarity-js -> local Ingest evidence:

- Mouse wire: `0.12345679104328156 / 1.23456789 / 7.89012345`
- Mouse PPG1: `0.1235 / 1.2346 / 7.8901`
- Touch wire: `0.7654321193695068 / 23.456789 / 34.567891`
- Touch PPG1: `0.7654 / 23.4568 / 34.5679`
- Pointer-bearing upload returned HTTP 204.

BrowserStack real-device touch matrix:

| Engine/device | Raw pointerdown geometry | Encoded TouchStart | Backend PPG1 | Result |
|---|---|---|---|---|
| Blink, Chrome, Pixel 7, Android 13 | `0.7 / 20 / 30` | `0.7 / 20 / 30` | `0.7 / 20 / 30` | PASS |
| Gecko, Firefox, Pixel 7, Android 13 | `0.7 / 52 / 78` | `0.7 / 52 / 78` | `0.7 / 52 / 78` | PASS |
| WebKit, Safari, iPhone 15, iOS 17 | `0 / 45.9957275 / 45.9957275` | same | `0 / 45.9957 / 45.9957` | PASS |

BrowserStack session IDs:

- Blink: `6e7ff006518c48302b35677da74cd4cd87fce819`
- Gecko: `d74a57f2ecee7e129e0f31a69b9a3a8d7b5aecf3`
- WebKit: `6fd420bfba2157acad52b23ea4b16f75dfbf4d18`

The mobile browsers also emit compatibility MouseDown events with unavailable geometry;
those retain the existing `-1` sentinel behavior and are separate from the validated
primary TouchStart record.

Final checks: pointer Playwright 5/5, real WebKit local smoke PASS, bundle delta +223
bytes (+0.2977%), Ingest build 0 warnings/errors, backend PPG1 test PASS.

packages/clarity-js/src/interaction/encode.ts
shawj0: Most PointerEvent events expose  pressure, width, and height :
•  pointerdown 
•  pointermove 
•  pointerrawupdate 
•  pointerup 
•  pointercancel 
•  pointerover  /  pointerenter 
•  pointerout  /  pointerleave 

The useful values are mainly during  pointerdown  and active  pointermove . On release or hover, pressure is normally 0 ; mouse width/height commonly default to 1.

packages/clarity-js/types/data.d.ts
```diff
@@ -78,6 +78,7 @@ export const enum Event {
     Focus = 50,
     CustomElement = 51,
     Chat = 52,
+    PointerContact = 53,
```
nairmanu: why PointerContact vs PointerDown?

packages/clarity-js/types/interaction.d.ts
```diff
@@ -120,6 +127,15 @@ export interface PointerData {
     isPrimary?: boolean;
 }
 
+export interface PointerContactData {
+    pointerType: PointerType;
```
nairmanu: nit: isnt type enough? do we have to call it pointerType?

packages/clarity-js/types/data.d.ts
shawj0: Agreed. Renamed event 53 from `PointerContact` to `PointerDown` in frontend commit `513f3f7` (with coordinated backend commit `3b8cd9b4ed`). The name now directly describes the browser event that produces it.

packages/clarity-js/types/interaction.d.ts
shawj0: Agreed. Renamed the decoded frontend field from `pointerType` to `type` in commit `513f3f7`, following the existing interaction-data convention. The positional wire field and backend `PointerType` model remain unchanged.

packages/clarity-js/src/interaction/down.ts
```diff
@@ -0,0 +1,18 @@
+import { Event } from "@clarity-types/data";
+import { PointerType } from "@clarity-types/interaction";
+import { queue } from "@src/data/upload";
+import { metadata } from "@src/layout/target";
+
+export default async function (
```
nairmanu: cant we keep this in the pointer file itself?

packages/clarity-js/src/layout/target.ts
```diff
@@ -13,19 +13,19 @@ export function target(evt: UIEvent): Node {
     return node && node.nodeType === Node.DOCUMENT_NODE ? (node as Document).documentElement : node;
 }
 
-export function metadata(node: Node, event: Event, text: string = null): TargetMetadata {
+export function metadata(node: Node, event: Event, text: string = null, track: boolean = true): TargetMetadata {
```
nairmanu: why this change?

packages/clarity-js/src/interaction/down.ts
shawj0: sure, moved

packages/clarity-js/src/interaction/encodeV2.ts
```diff
@@ -18,6 +19,20 @@ import * as unload from "@src/interaction/unload";
 import * as visibility from "@src/interaction/visibility";
 import * as focus from "@src/interaction/focus";
 
+export async function encodeDown(
```
nairmanu: this should be part of the current default function - not a standalone method

shawj0: Sure, done

packages/clarity-js/src/interaction/pointer.ts
```diff
@@ -10,13 +10,15 @@ import { target } from "@src/layout/target";
 import encode from "@src/interaction/encode";
 
 export let state: PointerState[] = [];
+export let down: PointerDownState[] = [];
```
nairmanu: merge down with state, extend state to have the additional fields, keep it optional for non down events

packages/clarity-js/src/interaction/encode.ts
```diff
@@ -22,6 +22,25 @@ export default async function (type: Event, ts: number = null): Promise<void> {
     let t = ts || time();
     let tokens: Token[] = [t, type];
     switch (type) {
+        case Event.PointerDown:
```
nairmanu, line 25: instead of a separate case statement, merge with other case statements, so that the for loop logic and common field logic is written only once. before queue call,  do additional if check for event type and push type pressure width and height

packages/clarity-js/src/interaction/pointer.ts
shawj0: Sure, merged with the existing pointer state and made the down-only fields optional.

packages/clarity-js/src/interaction/encode.ts
shawj0: Sure, merged into the shared pointer case and loop.

PR #1124, opened 2026-07-30
Add configurable restart delay(0.8.69)

nairmanu:
Makes the single-page-app restart delay configurable via a new optional `restart` config option, instead of the hardcoded `Setting.RestartDelay`.

When the URL changes on a SPA (history `pushState`/`replaceState`/`popstate`), `core/history.ts` stops Clarity and restarts it after a short delay. That delay was hardcoded to 250ms; it now comes from `config.restart`.

### DIFF
--- a/packages/clarity-js/types/core.d.ts
+++ b/packages/clarity-js/types/core.d.ts
@@ -142,6 +142,7 @@ export interface Config {
     includeSubdomains?: boolean;
     modules?: string[];
     diagnostics?: boolean;
+    restart?: number;
 }
 
 export const enum Constant {
--- a/lerna.json
+++ b/lerna.json
@@ -2,6 +2,6 @@
   "packages": [
     "packages/*"
   ],
-  "version": "0.8.68",
+  "version": "0.8.69",
   "npmClient": "yarn"
 }
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "clarity",
   "private": true,
-  "version": "0.8.68",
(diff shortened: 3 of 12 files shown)


nairmanu:
Thanks for the review! Addressed all three points in 37e207f:

1. **`{ restart: undefined }` could zero-out the delay** — `core/history.ts` now schedules the restart with `config.restart ?? Setting.RestartDelay`, so a caller passing `restart: undefined` (common with optional TS props / spreads) falls back to the 250ms default instead of restarting immediately via `setTimeout(0)`.
2. **Missing semicolon** — added the trailing `;` on the `import * as core from "@src/core"` line for consistency with the surrounding imports.
3. **No test coverage** — extended `test/history.test.ts` with two tests that stub `window.setTimeout` and record the delay:
   - asserts the configured `restart` delay is used to schedule the restart on a SPA URL change, and
   - a regression guard asserting `restart: undefined` falls back to 250ms (not 0).

All three history tests pass locally (`npx playwright test test/history.test.ts`).

ian7yang (approved):
LGTM

jsj85 (approved):

packages/clarity-js/types/core.d.ts
```diff
@@ -142,6 +142,7 @@ export interface Config {
     includeSubdomains?: boolean;
     modules?: string[];
     diagnostics?: boolean;
+    restart?: number;
```
shawj0, line 145: Is restartDelay clearer here?

shawj0 (approved):

PR #1139, opened 2026-09-10
Add support for pointer events (0.8.70)

shawj0:


packages/clarity-js/src/interaction/encodeV2.ts
```diff
@@ -40,9 +41,16 @@ export default async function (type: Event, ts: number = null): Promise<void> {
...
                     tokens.push(entry.data.isPrimary === undefined ? "true" : "" + entry.data.isPrimary);
-                    tokens.push(pTarget.region || Constant.Empty);
+                    if (entry.event === Event.PointerDown) {
+                        tokens.push(entry.data.type!);
+                        tokens.push(entry.data.pressure!);
+                        tokens.push(entry.data.width!);
+                        tokens.push(entry.data.height!);
+                    } else {
+                        tokens.push(pTarget.region || Constant.Empty);
+                    }
```
nairmanu, line 49: any issues with removing the else and push the region all the time

test/pointer.test.ts
```diff
@@ -0,0 +1,282 @@
...
+    expect(pointerDown[0].data.target).toBe(down.data.target);
+    expect(pointerDown[0].data.id).toBe(source.id);
+    expect(pointerDown[0].data.isPrimary).toBe(true);
+    expect(pointerDown[0].data.type).toBe(1);
+    expect(pointerDown[0].data.x).toBe(source.x);
+    expect(pointerDown[0].data.y).toBe(source.y);
+    expect(pointerDown[0].data.pressure).toBe(source.pressure);
+    expect(pointerDown[0].data.width).toBe(source.width);
+    expect(pointerDown[0].data.height).toBe(source.height);
+    expect(down).toBeTruthy();
```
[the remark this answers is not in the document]
shawj0: Fixed in f6311b0. The test now asserts that the legacy MouseDown exists before reading down.data, and pressure/width/height use 7-digit toBeCloseTo comparisons.

packages/clarity-js/src/interaction/encodeV2.ts
shawj0: No large issues. the backend requires 11 tokens for pointer down, and rejects if there's another token (ie region). We can update decode to handle 12 token to accommodate encodeV2's with region if we have traffic that's both diagnostics true and encode v2.

shawj0: I removed the else

packages/clarity-decode/src/clarity.ts
```diff
@@ -2,7 +2,7 @@ import { Data, version } from "clarity-js";
 import { BaselineEvent, CustomEvent, DecodedPayload, DecodedVersion, DimensionEvent } from "../types/data";
 import { LimitEvent, MetricEvent, PingEvent, SummaryEvent, UpgradeEvent, UploadEvent, VariableEvent, ExtractEvent, ConsentEvent } from "../types/data";
 import { FraudEvent, LogEvent, ScriptErrorEvent } from "../types/diagnostic";
-import { ChangeEvent, ClickEvent, ContextMenuEvent, ClipboardEvent, InputEvent, PointerEvent, ResizeEvent, ScrollEvent } from "../types/interaction";
+import { ChangeEvent, ClickEvent, ContextMenuEvent, ClipboardEvent, InputEvent, PointerEvent, PointerDownEvent, ResizeEvent, ScrollEvent } from "../types/interaction";
```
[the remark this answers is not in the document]
shawj0: Fixed in 22ab620 by splitting the interaction type imports so the changed clarity-decode source line stays below the enforced 140-character limit.

nairmanu (approved):

packages/clarity-js/src/interaction/encode.ts
```diff
@@ -40,8 +41,14 @@ export default async function (type: Event, ts: number = null): Promise<void> {
                     tokens.push(entry.data.y);
                     tokens.push(entry.data.id !== undefined ? entry.data.id : Constant.Empty);
                     tokens.push(entry.data.isPrimary === undefined ? "true" : "" + entry.data.isPrimary);
+                    if (entry.event === Event.PointerDown) {
+                        tokens.push(entry.data.type!);
```
rgaleev: **`PointerDownData` is declared but never enforced here.**

`pointer.state` holds `PointerData`, where these four fields are optional - hence the `!` assertions. If an entry ever reaches this branch without them, `undefined` is serialized positionally and the decoder reads `undefined`/`NaN` with no error, since the tokens are index-based.

The strict `PointerDownData` interface added in this PR would prevent exactly that, but it is only referenced from the decode types. A discriminated union on `pointer.state` would let the narrowing above drop these assertions.

packages/clarity-js/src/interaction/pointer.ts
```diff
@@ -25,26 +26,72 @@ export function observe(root: Node): void {
...
+            time: time(evt),
+            event: Event.PointerDown,
+            data: {
+                target: target(evt),
+                x,
+                y,
+                id: evt.pointerId,
+                isPrimary: evt.isPrimary,
+                type: pointerType,
+                pressure: evt.pressure,
```
rgaleev, line 52: **These three go on the wire as raw floats.**

Every other coordinate in this file is `Math.round()`ed (see `coordinates()` just below), but `pressure`/`width`/`height` are sent unrounded - and `test/pointer.test.ts` pins 7 decimal places via `toBeCloseTo(..., 7)`, which makes that precision part of the wire contract.

Is full float precision needed for diagnostics, or would `Math.round(pressure * 100)` do? Narrowing it later is a decoder-breaking change.

packages/clarity-js/src/interaction/pointer.ts
```diff
@@ -25,26 +26,72 @@ export function observe(root: Node): void {
     bind(root, "mousemove", mouse.bind(this, Event.MouseMove, root), true);
     bind(root, "wheel", mouse.bind(this, Event.MouseWheel, root), true);
     bind(root, "dblclick", mouse.bind(this, Event.DoubleClick, root), true);
+    if (config.diagnostics) {
```
rgaleev, line 29: **The gate is evaluated once, at bind time.**

`config.diagnostics` is read inside `observe()`, so toggling it after Clarity starts has no effect. Intended?

Also worth noting for rollout: `pointerdown` fires alongside `mousedown`/`touchstart`, so enabling this roughly doubles queued pointer events for those interactions.

packages/clarity-decode/src/clarity.ts
```diff
@@ -84,6 +85,10 @@ export function decode(input: string): DecodedPayload {
                 if (payload.upload === undefined) { payload.upload = []; }
                 payload.upload.push(data.decode(entry) as UploadEvent);
                 break;
+            case Data.Event.PointerDown:
```
rgaleev, line 88: **Consumers below 0.8.70 will log an error for every one of these.**

Unrecognized events fall through to `console.error(`No handler for Event: ...`)` at the bottom of this switch. Non-fatal, but the noise appears precisely when diagnostics is enabled. Worth confirming the backend decode path is on >=0.8.70 before enabling per-project.

test/helper.ts
```diff
@@ -4,6 +4,9 @@ import { resolve } from 'path';
 import type { Page } from '@playwright/test';
 import type { Core, Data, Layout } from "clarity-decode";
 
+const inputEvent = 27;
+const changeEvent = 42;
```
rgaleev: **nit:** these duplicate `Data.Event` values rather than deriving from it. If the enum shifts, the wait condition below silently stops matching and the helper hangs until timeout instead of failing clearly. Importing the enum would keep them in sync.

rgaleev (commented):
Reviewed locally against `refs/pull/1139/head`. Overall this looks solid - correctly gated behind the existing `config.diagnostics` (already `false` on master), token layout matches the decoder in both V1 and V2, versions are consistent at 0.8.70 across all seven manifests, and `PointerDown` is rightly excluded from `baseline.track` so activity is not double-counted.

Nice catch on the `x !== null` fix in `coordinates()` - the old truthiness check silently skipped the iframe offset when a coordinate was legitimately `0`, and it now has a dedicated test. The masking/helper test changes are genuine improvements too (order-independent lookup by type, deterministic event wait replacing a flaky `length > 2`).

Comments inline. Also: 40 commits with a few empty messages - worth squashing on merge.

packages/clarity-js/src/interaction/encode.ts
shawj0: Fixed in 9be5dbd. `PointerState` is now discriminated by `event`: Event 53 requires `PointerDownData`, while legacy pointer events use `PointerData`. Both encoders narrow without non-null assertions, with no wire-format change.

packages/clarity-js/src/interaction/pointer.ts
shawj0: The browser-provided precision is intentional.

packages/clarity-js/src/interaction/pointer.ts
shawj0: Yes, both behaviors are intentional. Diagnostics is startup configuration, so the listener gate is evaluated when observation is initialized. During migration, Event 53 is collected alongside mousedown/touchstart so we can compare coverage and performance without removing established telemetry; it adds press work only, not duplicate move collection. I added this explicitly to the PR description and rollout risk.

packages/clarity-decode/src/clarity.ts
shawj0: Confirmed. `clarity-decode` 0.8.70 handles Event 53; older JavaScript decoder packages follow their existing non-fatal unknown-event path. The coordinated backend decoder support must be deployed before the project flight is enabled. The PR description now calls out this deployment order explicitly.

test/helper.ts
shawj0: Fixed in 9be5dbd. The helper now derives both values from `Data.Event.Input` and `Data.Event.Change`, so enum changes fail at compile/build time rather than timing out in the wait condition.

test/helper.ts
```diff
@@ -2,7 +2,8 @@ import { readFileSync } from 'fs';
 import { pathToFileURL } from 'url';
 import { resolve } from 'path';
 import type { Page } from '@playwright/test';
-import type { Core, Data, Layout } from "clarity-decode";
+import { Data } from "clarity-decode";
```
rgaleev: **Blocker: this breaks `markup()` at runtime — every test that calls it will fail.**

`Data.Event` is a `const enum` declared only in `clarity-js/types/data.d.ts`. It has no runtime representation, and `clarity-decode`'s runtime entry (`src/index.ts`) exports only `decode` — the `Data` namespace exists solely in `types/index.d.ts`.

Verified against the built bundle:

```
$ node -e "const m=require('clarity-decode'); console.log(Object.keys(m), typeof m.Data)"
[ 'decode' ] undefined
```

Playwright transpiles tests with Babel, which does not inline const enums across modules, so the import is kept and `Data` resolves to `undefined`. Confirmed with a probe spec on this commit:

```
PROBE typeof Data = undefined
PROBE THREW: Cannot read properties of undefined (reading 'Event')
```

Line 39 (`{ input: Data.Event.Input, change: Data.Event.Change }`) throws before `waitForFunction` is reached.

Sorry — my nit asked for exactly this and I should have flagged the const-enum constraint with it. Options that keep the intent without the runtime import:

- Re-declare locally with a type check that still fails compilation on drift:
  ```ts
  import type { Data } from "clarity-decode";
  const inputEvent: Data.Event.Input = 27;
  const changeEvent: Data.Event.Change = 42;
  ```
  Typed to the enum member, so changing the enum breaks the build rather than hanging the test.
- Or revert to the plain constants from the previous commit — the nit is not worth a runtime dependency.

Worth confirming `npx playwright test test/masking.test.ts` passes before merge; I believe it currently does not.

rgaleev (changes requested):
Re-reviewed at `9be5dbd`.

**The discriminated union is exactly right.** `PointerEventState` / `PointerDownState` split removes all four `!` assertions and makes the `PointerDown` branch provably total. I checked out the commit and ran the repo's pinned `tsc` (4.9.5) over `clarity-js` — **zero type errors** in `pointer.ts`, `encode.ts`, `encodeV2.ts`, `interaction.d.ts`. Typing `mouse()`/`touch()` as `PointerEventType` is a nice tightening beyond what I suggested.

**One blocker introduced by the `test/helper.ts` change** — details inline. My nit was the proximate cause, and I should have flagged the const-enum constraint when I raised it.

Still open from the previous round (neither is blocking):
- **`pressure`/`width`/`height` still ship as raw floats**, with `toBeCloseTo(..., 7)` pinning 7 decimals into the wire contract. Fine if deliberate — worth an explicit decision, since narrowing later is decoder-breaking.
- **Diagnostics gate is still read once at bind time.** Almost certainly fine; flagging only so it is a choice rather than an accident.

rgaleev:
**Correction to my last review: the two "still open" items were already answered — my mistake.**

@shawj0 had replied to both before I posted (browser-provided precision is intentional; the diagnostics gate and dual collection are deliberate and now documented in the description). I diffed the code, saw no change, and wrote "not addressed" without reading the replies — but unchanged code is not the same as an unanswered question, and both were answered clearly. Those threads are correctly resolved.

The `test/helper.ts` const-enum blocker is the only thing outstanding from me.

test/helper.ts
shawj0: Fixed in 90dba10. `Data` is now imported type-only, and the runtime values are local constants typed as the corresponding enum members. This preserves compile-time drift detection without relying on a nonexistent `clarity-decode` runtime export. `yarn test test/masking.test.ts` passes 4/4.

test/helper.ts
```diff
@@ -4,6 +4,9 @@ import { resolve } from 'path';
 import type { Page } from '@playwright/test';
 import type { Core, Data, Layout } from "clarity-decode";
 
+const inputEvent: Data.Event.Input = 27;
```
rgaleev, line 7: **My suggestion was wrong on one point - this annotation does not actually catch drift.**

The runtime fix is correct and `markup()` works again. But a numeric literal is assignable to a numeric enum type in TypeScript, so the annotation compiles with *any* number:

``````ts
const a: Data.Event.Input = 99;   // compiles cleanly - no error
``````

I verified this against the repo's own TypeScript, with the import resolving (a string literal in the same position *does* error, so `Data.Event.Input` is being read correctly). So if `Input` ever moves off 27, this still compiles and the helper goes back to hanging until timeout - the exact failure the constants were meant to prevent.

If you want the guard to actually bite, a type-level equality assertion does work (verified both directions - wrong value errors, correct value compiles):

``````ts
type AssertEq<X, Y> = [X] extends [Y] ? ([Y] extends [X] ? true : never) : never;

const inputEvent = 27;
const changeEvent = 42;
const _assertInput: AssertEq<typeof inputEvent, Data.Event.Input> = true;
const _assertChange: AssertEq<typeof changeEvent, Data.Event.Change> = true;
``````

Entirely your call whether that is worth the ceremony for two test constants - a plain comment pointing at `Data.Event` would also be honest. I would just rather not leave a guard in place that reads as protective but is not. Not blocking either way.

rgaleev (approved):
Re-reviewed at `90dba10`. **The blocker is fixed** - dropping the value import resolves it, and I confirmed the previously-broken suite now passes:

``````
npx playwright test test/masking.test.ts
  4 passed (7.7s)
``````

That was throwing `Cannot read properties of undefined (reading 'Event')` before this commit, so `markup()` and everything downstream of it is healthy again.

Approving. One follow-up inline about the typed constants: the pattern I suggested does not do what I claimed, and I would rather say so than leave it. It is a nit, not a blocker.

Everything else from my earlier reviews still stands as resolved: the discriminated union is clean, the float precision and the bind-time gate were both answered and are intentional.

rgaleev (approved):

PR #1142, opened 2026-09-15
Performant Agentic Browser Detection

shawj0:
Adds performant Claude and Codex browser presence detection, including Codex overlay and sidebar-comment roots, with explicit per-product deduplication.


packages/clarity-devtools/static/manifest.json
```diff
@@ -2,8 +2,8 @@
   "manifest_version": 3,
   "name": "Microsoft Clarity Developer Tools",
   "description": "Clarity helps you understand how users are interacting with your website.",
-  "version": "0.8.70",
-  "version_name": "0.8.70",
+  "version": "0.8.71",
```
nairmanu: why is version 71 here?

packages/clarity-js/src/layout/agent.ts
```diff
@@ -0,0 +1,73 @@
...
+const enum AgenticBrowser {
+    None = 0,
+    Claude = 1,
+    Codex = 2
+}
+
+export function start(): void {
+    seen = [false, false, false];
+    scan();
+}
```
nairmanu: cant we just maintain a Set() and enter Claude and Codex into it when seen?

packages/clarity-devtools/static/manifest.json
shawj0: fixed, thanks

packages/clarity-js/src/layout/agent.ts
shawj0: sure, done

packages/clarity-js/src/layout/agent.ts
```diff
@@ -0,0 +1,85 @@
...
+    let element = node as HTMLElement;
+    let signal = identify(element.id, parent || element.parentElement);
+    let browser = classify(signal);
+    if (browser !== AgenticBrowser.None && !seen.has(browser)) {
+        seen.add(browser);
+        dimension.log(Dimension.AgenticBrowserSignal, signal.toString());
+    }
+}
+
+function find(id: string, parent: HTMLElement): HTMLElement {
```
nairmanu: isnt this scanning every single child of body and document element on initial clarity load? arent we already doing that as part of discover? this seems we are duplicating the scan

shawj0: Alright, now it's detecting in three places. 1st in the discover traversal like you're saying, 2nd in the child list mutations, and 3rd at window.load when delay dom defers the mutation observer

nairmanu (approved):

